PR #8, opened 2016-08-03
Get progress from `request.upload.onprogress` as well.

ryanto:
In order to get upload progress the handler must be attached to
`request.upload.onprogress`.

See:
https://developer.mozilla.org/en-US/docs/Using_files_from_web_applications


ryanto:
By the way thanks for this addon, it makes uploading so easy!

Issue #9, opened 2016-09-21
Support for application/octet-stream uploads and file chunking


jlevycpa:
This looks like a great addon. I like the approach of using a service a lot. For me to use this in my project I would need two things. First, I need to support sending uploads without `multipart/form-data` encoding. Basically I would just put the blob content in the request body and use `application/octet-stream` as the content type. 

Would you be interested in a PR to add this as an option? Maybe `options.encoding` with options `form` (default) or `octet`?

Second, I'd like to see an option to upload large files using chunking. One way to achieve this would be to depend on something like [resumable.js](http://www.resumablejs.com/). Is this a direction you are interested in?

Please let me know your thoughts. Thanks!

tim-evans:
Yeah, that'd be awesome!

jelhan:
Support for `application/octet-stream` should be added by #12. Upload in chunks is not supported yet as far as I know. This would either add a lot of additional complexity or require another dependency to be pushed to all consumers. To be honest I'm not sure if this is reasonable for a feature which is only needed for a few special cases. Maybe this is more something for a specialized `ember-file-upload-resumable` addon? Tending towards closing this one as _won't fix_.

fsmanuel:
I implemented a way to upload the files in chunks. It needs some minor tweeks in the `upload` and `normalizeOptions` functions. Let me know what you think and if it's worth to open a PR.

```js
export default function uploadChunks(file, url, opts) {
  let size = get(file, 'size');
  let type = get(file, 'type');

  // Chunk
  let chunkStart = opts.chunkStart || 0;
  let chunkEnd = Math.min(chunkStart + opts.chunkSize, size);
  let chunk = get(file, 'blob').slice(chunkStart, chunkEnd, type);

  let filename = get(file, 'name');
  let filenameEncoded = encodeURI(filename)

  // Headers
  opts.headers = {};
  opts.headers['Content-Range'] = `bytes ${chunkStart}-${chunkEnd-1}/${size}`;
  opts.headers['Content-Disposition'] = `attachment; filename="${filenameEncoded}"`;

  // Upload
  return upload(file, url, opts, (request, options) => {
      // Build the form
      let form = new FormData();

      Object.keys(options.data).forEach((key) => {
        if (key === options.fileKey) {
          form.append(key, chunk, filename);
        } else {
          form.append(key, options.data[key]);
        }
      });

      return request.send(form);
    })
    .then((response) => {
      // Progress
[...]

PR #12, opened 2016-10-04
Add payloadType option to file

jlevycpa:
This adds the option to specify `payloadType: binary` for a file, which uploads the blob as `application/octet-stream` without any `multipart/form-data` encoding.

As discussed in #9.


jlevycpa:
Some testing is probably in order to make sure this doesn't break any existing functionality. I don't have a setup to test the multipart/form-data method.

tim-evans:
This is interesting! I'm not sure if this does everything, since it should pass along all the data as well as the payload.

jlevycpa:
Good point. The only way to pass additional data would be query parameters. That's what [this library](https://github.com/23/resumable.js/pull/49/commits/620ed98b1763d7093d9e010131e6c17c8c438ca5) did when they added support for this. Personally, I'm just using HTTP headers which works fine. 

I agree that at a minimum, the docs should specify that the data option is ignored in binary mode. Thoughts?

tim-evans:
HTTP headers sound good to me!

jlevycpa:
Do you mean that the library should set the contents of `data` as HTTP headers? Or the docs should advise to use the `headers` option instead of `data`? My vote would be for less magic, and to just document that `data` is ignored.

tim-evans:
I think that'd make the most sense to use headers.

ef4:
Thanks, this is a nice feature. I extended it slightly so that it's still possible to pick the content type. `application/octet-stream` is not the only thing one might want to directly submit. I'm using this with s3, where it's nice to just PUT any file.

tim-evans:
@ef4 I can merge this, would you like to add your changes after?

ef4:
Looks like you have already merged them in https://github.com/tim-evans/ember-file-upload/pull/23

Thanks, looks good.

maxigs:
Is it possible that s3 upload does not work without this binary option?
I've been trying it out and the file on s3 has content type of form data and is not usable that way.

jlevycpa:
Yes, I believe that is the case. The S3 API expects the payload of the request to be just the binary file. It's not expecting form encoding.

maxigs:
Thanks @jlevycpa.

I'm not sure what the right way is of helping to get this merge request into master, so i made a little fork of the current and added your logic to it, so i could try and use it.
Seems to work fine so far: https://github.com/maxigs/ember-file-upload

Please let me know if i can help to get it into a proper released version :)

tim-evans:
@maxigs I think I'm trying to understand the ramifications of this. How does extra data get passed along in this use case?

maxigs:
In the upload itself there is no way of adding extra data as it's just uploading the file itself to s3.
I'm not sure if there is way to attach any additional data to be stored at s3, i'm pretty new to using it.

tim-evans:
Here's the documentation around the headers that are sendable for that:

http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html

alexnj:
Thanks so much for this library. Wish it featured an option to set `Content-Type` instead. Unfortunately, hardcoding content type to `application/octet-stream` as proposed in this PR wouldn't do. S3 requires the content type hashed in a signed request to match that of the actual upload request. I.e., if it was `Content-Type: image/png`, the upload needs to be same.

bissleb:
Including a big +1 on getting this merged.  The use case this would primarily support is uploading to S3 via pre-signed urls. 

http://docs.aws.amazon.com/AmazonS3/latest/dev/PresignedUrlUploadObject.html

Using these, the Content-Type on the request needs to match what was provided when creating the PresignedUrl. In this case, "application/octet-stream" is entirely appropriate, as it represents a stream of data to be saved to a file defined in the PresignedUrl. The ability to specify the Content-Type header explicitly would also work, of course. But really, S3 doesn't care.

I have tried it out, and maxigs's implementation works great. Without it, I can't use this great (seriously, thanks for the hard work) module.

alexnj:
> In this case, "application/octet-stream" is entirely appropriate

@bissleb It seems you are ignoring the original content types of objects in your S3 bucket. However, that wouldn't work for applications that require original `Content-Type` to be stored (For example, CloudFront serving image assets with S3 buckets as origin. See http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AddingObjects.html). 

I might have [a solution to this](https://github.com/tim-evans/ember-file-upload/compare/master...alexnj:master). I will spend some time to raise a PR.

tim-evans:
Thanks @alexnj

bissleb:
Yes, that's true, @alexnj. We are copying the files from the upload bucket to another and setting the appropriate metadata on the new object. So, probably a narrower use case than other people.

Thanks! Let me know if I can help at all.

queenvictoria:
Hi there. Once the version from @maxigs is in, and the mime-type option from @alexnj, I can update the `README` with a Node JS getSignedUrl pattern and the updated URL pattern. I've got the thing working but it was a long way around.

In particular (this bit took the most time) you need to set the method to 'PUT'.

```javascript
        return file.upload(response.url, {
          data: response.credentials,
          method: 'PUT',
          payloadType: 'binary',
          contentType: file.blob.type,
        });
```

If I get a moment I'll PUT all this together in a single PR.

tim-evans:
@queenvictoria that'd be awesome. I'll get to merging this soon. Sorry for delays on this; it's a large feature that I want to be sure about the API with.

tim-evans:
Hey all, would you be fine if I made a new function called `uploadBinary` or `uploadRaw` instead of having a payloadType?

queenvictoria:
Fine with me!

queenvictoria:
So for future readers: although this commit https://github.com/tim-evans/ember-file-upload/commit/f2f01b91d7214174146dd41978894db30db7781c looks like it supports `payloadType`, if you're installing 2.0.0+ then as @tim-evans says, there is a new method `uploadBinary`.

So this is what was happening with < 2.0.0:
```javascript
        return file.upload(response.url, {
          payloadType: 'binary',
          method: 'PUT',
          contentType: file.blob.type,
          data: JSON.stringify(response.credentials),
        });
```

And this is what is happening now:
```javascript
        return file.uploadBinary(response.url, {
          method: 'PUT',
          contentType: file.blob.type,
          data: JSON.stringify(response.credentials),
        });
```

So good news! It all works out of the box.

queenvictoria:
So for future readers: although this commit https://github.com/tim-evans/ember-file-upload/commit/f2f01b91d7214174146dd41978894db30db7781c looks like it supports `payloadType`, if you're installing 2.0.0+ then as @tim-evans says, there is a new method `uploadBinary`.

So this is what was happening with < 2.0.0:
```javascript
        return file.upload(response.url, {
          payloadType: 'binary',
          method: 'PUT',
          contentType: file.blob.type,
          data: JSON.stringify(response.credentials),
        });
```

And this is what is happening now:
```javascript
        return file.uploadBinary(response.url, {
          method: 'PUT',
          contentType: file.blob.type,
          data: JSON.stringify(response.credentials),
        });
```

So good news! It all works out of the box.

tim-evans:
Yay! Glad to hear!!

queenvictoria:
Hey @tim-evans I'd like to write up a S3, NodeJS/Express and Ember pattern. I see you're moving away from the README. So I can't make a PR correct? Anywhere I can do it more sensibly?

tim-evans:
Feel free to add to the README; I'll migrate what you've written to the recipes in the docs app.

The docs are the dummy app in tests/dummy/app, FYI

PR #15, opened 2016-10-13
Fix #14 (Maximum call stack size exceeded)

### LINKED ISSUE #14: Typo in flushFilesWhenSettled and getting 'Maximum call stack size exceeded'
Think `file.status` should be `file.state` [here](https://github.com/tim-evans/ember-file-upload/blob/master/addon/services/file-queue.js#L78)?

Also, `set(this, 'files', Ember.A())` gives me `RangeError: Maximum call stack size exceeded`

Padchi:
Well first of, this is my first fork and pull request so don't know if i'm doing things right..

These changes is not tested but works on my machine and stop the error explained in #14.

### DIFF
--- a/addon/queue.js
+++ b/addon/queue.js
@@ -148,6 +148,10 @@ export default Ember.Object.extend({
     let allFilesHaveSettled = files.every(function (file) {
       return ['uploaded', 'aborted'].indexOf(file.state) !== -1;
     });
+    
+    if (files.length <= 0) {
+      return;
+    }
 
     if (allFilesHaveSettled) {
       get(this, 'files').forEach((file) => set(file, 'queue', null));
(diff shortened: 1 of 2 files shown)


addon/queue.js
```diff
@@ -148,6 +148,10 @@ export default Ember.Object.extend({
     let allFilesHaveSettled = files.every(function (file) {
       return ['uploaded', 'aborted'].indexOf(file.state) !== -1;
     });
+    
+    if (files.length <= 0) {
```
tim-evans: Could we change this to `files.length === 0`?

I don't know how we could have an array with less than 0 items in it.

tim-evans:
You're doing great! This is ok for this, since I need to flesh out the tests 😄

Padchi:
Good! And of course can you change the syntax 😉 Have some more bugs but I want to test it first so it's not me doing anything wrong before i submit a issue.
You want me to write tests before submit another pull?

tim-evans:
I can do that, go ahead and submit a few more bug reports!

GreatWizard:
Thanks for this fix, I just start a new project with ember-file-upload and I don't understand what was the problems before saw this fix! Thanks @Padchi.

Is it possible to have a new patch release ? :)
Thanks @tim-evans for this pretty nice addon!

tim-evans:
Released beta 19

PR #22, opened 2017-02-03
Fixes #21 Drag-and-drop isn't working with IE11

### LINKED ISSUE #21: Drag-and-drop isn't working with IE11 (and let's face it, probably other versions of IE)
When dropping a file from Windows Explorer into IE11, I get this error in the console: `SCRIPT87: Invalid argument.`.

It then proceeds to highlight [this line](https://github.com/tim-evans/ember-file-upload/blob/85b48afd2694f5e95fd6d8590b16910d09ef5e24/addon/components/file-dropzone/component.js#L121) from `file-dropzone/component.js`:

```javascript
let url = evt.dataTransfer.getData('text/uri-list');
```

I've scoured Google looking for a way to help fix this. I tried changing `text/uri-list` to `URL` [per MSDN](https://msdn.microsoft.com/en-us/library/ms536436(v=vs.85).aspx) with no success. (`console.log(url)` ends up evaluating to `undefined`.)

I'm not sure if you're interested in supporting IE or if this even works in Edge, but I thought that I would log what I've run into here if anyone was interested in fixing it.

chrisdpeters:
This silences the IE errors mentioned in #21.

I'm still learning the JavaScripts, so let me know if something could be written better or tested or something.

### DIFF
--- a/addon/components/file-dropzone/component.js
+++ b/addon/components/file-dropzone/component.js
@@ -14,6 +14,9 @@ let supported = (function () {
          'draggable' in document.createElement('span');
 }());
 
+let supportsHtml = null;
+let supportsUrls = null;
+
 export default Ember.Component.extend({
 
   layout,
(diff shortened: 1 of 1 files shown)


addon/components/file-dropzone/component.js
```diff
@@ -118,16 +118,30 @@ export default Ember.Component.extend({
...
     // Testing support for dragging and dropping images
     // from other browser windows
-    let url = evt.dataTransfer.getData('text/uri-list');
-    let html = evt.dataTransfer.getData('text/html');
-    if (html) {
+    let supportsUrls, supportsHtml, html, url;
+
+    try {
+      html = evt.dataTransfer.getData('text/html');
+      supportsHtml = true;
```
chrisdpeters: @tim-evans In hindsight, I guess it doesn't even really make much sense to store values like `supportsHtml` and `supportsUrls` the way that I'm doing it here. The `if` tests below would still receive `undefined` `html` and `url` values.

I know in #21, you had mentioned storing `supportsHtml` and `supportsUrls` properties on the component (e.g., `this.set('supportsHtml, true)`), but I wasn't sure how you'd want to use those in context of the component.

tim-evans: This looks good; caching this globally may be nice 😄

tim-evans: (at the top of the file)

chrisdpeters: Would that even be possible if you need an event object to begin with? I'm learning a lot here, so I'd appreciate any insight that you have.

tim-evans: Sure! I understand where you're going with it, but I'm thinking something closer to moving the `let` to the top of the file and only checking the first time we get into `drop`.

```js
let supportsHtml = null;

// ... later
if (supportsHtml == null) {
  try {
    supportsHtml = true;
  } catch () {
    supportsHtml = false;
  }
}
```

chrisdpeters: Makes sense! I thought you were saying to cache the actual logic at the top of the file, which left me scratching my head. (I spend roughly 50% of my day scratching my head, so this isn't out of the norm.)

See new commit 48a826d5ae12509dc4dd3d6693d75646497e04bc, which I agree is great for performance.

PR #23, opened 2017-02-05
Fix fastboot guard for IE11

ef4:
https://github.com/tim-evans/ember-file-upload/pull/22 makes drag & drop work for me on IE11. But `dropzone.supported` is incorrectly calculated as `false`.

This happens because on IE11, `window.hasOwnProperty('document')` is false.


ef4:
On second thought, I'm not sure #22 really fixed drag & drop, so maybe it's correct that `supported` is false.

ef4:
Ah, it's only multi-file drag & drop that is not working in my case. So this is probably still a good change.

tim-evans:
Sounds good! Thanks

PR #28, opened 2017-03-29
Permit setting withCredentials on the XMLHttpRequest for file-upload

imipolexg:
Right now there is no way that I can tell of setting `withCredentials: true` on the file upload, making it difficult to pass a session cookie through for authentication. This commit allows you to set `withCredentials: true` on the `options` object when calling `file.upload`.

### DIFF
--- a/addon/file.js
+++ b/addon/file.js
@@ -45,6 +45,8 @@ function normalizeOptions(file, url, options) {
 
   options.data[options.fileKey] = file.blob;
 
+  options.withCredentials = options.withCredentials || false;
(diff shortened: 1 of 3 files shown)


tim-evans:
Looks good! Could you add a test here: https://github.com/tim-evans/ember-file-upload/blob/master/tests/unit/system/http-request-test.js

imipolexg:
Added some basic tests. Let me know if you want something more than that.

tim-evans:
:heart:

PR #32, opened 2017-04-19
nullify _listener after file drop

EnotPoloskun:
Fix for issue 
https://github.com/tim-evans/ember-file-upload/issues/31


tim-evans:
This actually looks good!

Issue #36, opened 2017-05-11
`ember test` arrow.svg error (Windows?)


XaserAcheron:
I can't seem to get `ember test` to run. It fails with the following error when trying to clean up after building:

```
C:\Code\GitHub\forks\ember-file-upload>ember test
DEPRECATION: Addon test support files were detected in `C:/Code/GitHub/forks/ember-file-upload/node_modules/ember-cli-test-loader/addon-test-support`, but no JavaScript preprocessors were found for `ember-cli-test-loader`. Please make sure to add a preprocessor (most likely `ember-cli-babel`) to `dependencies` (NOT `devDependencies`) in `ember-cli-test-loader`'s `package.json`.
cleaning up...
Build failed.
arrow.svg
The Broccoli Plugin: [SVGOFilter] failed with:Error: ENOENT: no such file or directory, open 'C:\Users\XaserAcheron\AppData\Local\Temp\if-you-need-to-delete-this-open-an-issue-async-disk-cache\d0a26e6a212dbbaf16034028ed97fb47dc71c26b\{"persist":true,"svgoConfig":{}}337a3cdfc8b191f654659e3ce12929b1'
    at Error (native)

The broccoli plugin was instantiated at:
    at SVGOFilter.Plugin (C:\Code\GitHub\forks\ember-file-upload\node_modules\broccoli-persistent-filter\node_modules\broccoli-plugin\index.js:7:31)
    at SVGOFilter.Filter [as constructor] (C:\Code\GitHub\forks\ember-file-upload\node_modules\broccoli-persistent-filter\index.js:60:10)
    at new SVGOFilter (C:\Code\GitHub\forks\ember-file-upload\node_modules\broccoli-svg-optimizer\index.js:15:10)
    at Class.<anonymous> (C:\Code\GitHub\forks\ember-file-upload\node_modules\ember-svg-jar\lib\index.js:149:12)
    at Class.memoized [as optimizedSvgsFor] (C:\Code\GitHub\forks\ember-file-upload\node_modules\lodash\lodash.js:10551:27)
    at Class.<anonymous> (C:\Code\GitHub\forks\ember-file-upload\node_modules\ember-svg-jar\lib\index.js:158:50)
    at Class.memoized [as svgsFor] (C:\Code\GitHub\forks\ember-file-upload\node_modules\lodash\lodash.js:10551:27)
[...]

PR #37, opened 2017-05-11
'disabled' attribute for file-upload

### LINKED ISSUE #35: Ability to disable button & file input button
I've got a requirement to lock down a form from editing until the user does some stuff, and I'm having to resort to hacks to disable the dropzone and file input button. It'd to be handy to have proper disablement (is that a word?) as a supported feature.

XaserAcheron:
Closes #35.

This PR adds a new file-upload attribute, `disabled`, that if set, does two things:
- Prevents files from being added to the queue*
- Disables the `input` control (the upload button/link/whatnot).

*As a note, the queue can still be manually manipulated in code, e.g. `queue.push(file)`; this only affects the built-in behavior (i.e. `queue._addFile`, which the dropzone uses).

No tests, unfortunately (because of #36), but I did `yarn link` it to my own Ember app and ensure it works as expected. Docs included, at least.

Additional stuff like applying CSS styles is left as an exercise for the user. This just takes care of the core bits that can't be handled in a higher-level custom component.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ exposes a variety of parameters for configuring file-uploader:
 |---------------------|------------------|
 | `accept`            | a list of MIME types / extensions to accept by the input
 | `multiple`          | whether multiple files can be selected
+| `disabled`          | if set, disables input and prevents files from being added to the queue
 | `onfileadd`         | the name of the action to be called when a file is added to a queue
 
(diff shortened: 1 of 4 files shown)


tim-evans:
LGTM! I'll need to add a test to this later 👍

PR #46, opened 2017-07-16
fix(file): file.upload() returns a promise

shuriu:
`file.upload()` broke in https://github.com/tim-evans/ember-file-upload/commit/aabc6e8231549ba636bad13c3770d72a8609daaf and was returning `undefined` instead of a promise.


tim-evans:
Good catch 😓

PR #47, opened 2017-07-16
feat(http-request): add jsonapi content type

shuriu:
Added `application/vnd.api+json` to the list of parsable JSON responses to make jsonapi servers happy 😄 .


tim-evans:
Looks good! Should this parse this as well into a format that's normalized, or should this remain as?

shuriu:
I think it should remain as is. The nuisance for me was that the `response.body` was a string because this content type wasn't covered in the condition to `JSON.parse`.

tim-evans:
Sounds good to me!

Issue #41, opened 2017-06-05
Multiple uploads of the same file


lodrantl:
Using the file-upload component, you can not upload the same file multiple times in a row, because the form is not cleared after file upload.
I know it sounds weird, but my latest application needs this use case. I have solved the problem by reopening the component and resetting the form like this.

```javascript
FileUpload.reopen({
  actions: {
    change(files) {
      this._super(files);
      this.$().children("input").val(null);
    }
  }
});
```
Do you think this could become a configurable attribute to the component or do you think it would cause some problems along the way?

tim-evans:
What you did should be ok to do- we already have a reference to the file then 👍 

Could you make a PR for this? With tests would be ✨ amaaaazing:sparkles:

lodrantl:
Will do later in the afternoon or tomorrow.

mickeysanchez:
Is this fixed?

tim-evans:
I don't think so.

matthewdias:
this issue can be closed now, yeah?

lodrantl:
I totally forgot about this... Thanks @mickeysanchez for the PR I promised. I'm closing this.

tim-evans:
Thanks @matthewdias 👍

PR #48, opened 2017-07-17
Add ability to upload same file twice in a row

mickeysanchez:
Fixes issue #41 

We tried writing an integration test for the component, but we couldn't find a good way to test this because:

- You can't set the value of a file input programmatically for security reasons - so we can't test that we're setting the value. 

- This is a symptom of the browser not triggering an event from user input, so simulated events do not replicate the issue - so we don't have a way to verify that this allows the event to be triggered a second time.


tim-evans:
Very fair points! Thanks for writing up an explanation for this!

mickeysanchez:
👍🏽

PR #49, opened 2017-07-20
Update README.md - fix integration example

bartocc:
photo was not defined in the catch block in the integration paragraph


tim-evans:
Thanks!

Issue #43, opened 2017-06-13
S3 Uploads without FormData


datajohnny:
I have a question how can I just so a single image upload without having a formdata wrapper.

ryanjavens:
@datajohnny do you have a code example to help your question?  I'd love to take a look and try to help you out!

datajohnny:
@ryanjavens Hi I was wondering if it was possible to just do the image itself like

```javascript
let $form = $('#file_upload');
let file = $form.files[0];
$.ajax(url, file, 'PUT');
```

donaldwasserman:
@datajohnny if you're still having some problems with that and s3, i think the problem is more on S3's side, which requires a bunch of different form data.

PR #54, opened 2017-08-14
[WIP] Use uploader objects

Subtletree:
Currently it's not trivial to customize how a file is uploaded.

**A scenario:**
I'm testing the new [activestorage](https://github.com/rails/rails/tree/master/activestorage) module in edge rails 5.2 for direct uploads.
It sends the client a signed s3 upload URL with a `Content-MD5` header of the expected data's checksum. That data is just the file itself, **not form data**.

Therefore I need to override `ember-file-upload` to send just the file and not wrap it in a form.

This PR makes customizations like this easy by moving the uploading logic from the `file` model to 'uploader' objects, in the same style as ember-data's serializers/adapters or ember-simple-auth's authenticators.

3 uploaders are included by default:

- BaseUploader
- FormDataUploader (default)
- BinaryUploader

To solve the above scenario I could create an `uploaders/application.js` file which would be used by default:
```
import BaseUploader from 'ember-file-upload/uploaders/base';
export default BaseUploader;
```
Or pass the uploader in when uploading the file:
```
file.upload(url, {
  method: 'PUT',
  headers: headers
}, 'uploader:base');
```
Or create a custom uploader to do whatever I want:
```
import BaseUploader from 'ember-file-upload/uploaders/base';

export default BaseUploader.extend({
  normalizeOptions(file, url, options) {
    options = this._super(...arguments);
    // do something with options
    return options;
  },

  createData(file, url, options) {
    // customize data
    return data;
  }
});
```

I'm using this fork in production currently.

It needs some polishing and documenting but wanted to gauge interest in moving in this direction 🙂 

These issues would be easily solved #9, #43

### DIFF
--- a/addon/file.js
+++ b/addon/file.js
@@ -1,116 +1,12 @@
 /* global atob, Uint8Array */
 import Ember from 'ember';
 import FileReader from './system/file-reader';
-import HTTPRequest from './system/http-request';
-import RSVP from 'rsvp';
 import uuid from './system/uuid';
 
 import get from 'ember-metal/get';
-import set from 'ember-metal/set';
 
 const { computed } = Ember;
 const { reads } = computed;
-
-function normalizeOptions(file, url, options) {
-  if (typeof url === 'object') {
-    options = url;
-    url = null;
-  }
-
-  options = options || {};
-
-  options.url = options.url || url;
-  options.method = options.method || 'POST';
-  options.accepts = options.accepts || ['application/json', 'text/javascript'];
-  if (!options.hasOwnProperty('contentType')) {
-    options.contentType = get(file, 'type');
-  }
-  options.headers = options.headers || {};
-  options.data = options.data || {};
-  options.fileKey = options.fileKey || 'file';
-
-  if (options.headers.Accept == null) {
-    if (!Array.isArray(options.accepts)) {
-      options.accepts = [options.accepts];
-    }
-    options.headers.Accept = options.accepts.join(',');
-  }
-
-  // Set Content-Type in the data payload
-  // instead of the headers, since the header
-  // for Content-Type will always be multipart/form-data
-  if (options.contentType) {
-    options.data['Content-Type'] = options.contentType;
-  }
-
-  options.data[options.fileKey] = file.blob;
-
-  options.withCredentials = options.withCredentials || false;
-
-  return options;
-}
-
-function upload(file, url, opts, uploadFn) {
-  if (['queued', 'failed', 'timed_out'].indexOf(get(file, 'state')) === -1) {
-    Ember.assert(`The file ${file.id} is in the state "${get(file, 'state')}" and cannot be requeued.`);
-  }
-
-  let options = normalizeOptions(file, url, opts);
-
-  let request = new HTTPRequest({
-    withCredentials: options.withCredentials,
-    label: `${options.method} ${get(file, 'name') } to ${options.url}`
-  });
-
-  request.open(options.method, options.url);
-
-  Object.keys(options.headers).forEach(function (key) {
-    request.setRequestHeader(key, options.headers[key]);
-  });
-
-  if (options.timeout) {
-    request.timeout = options.timeout;
-  }
-
-  request.onprogress = function (evt) {
-    if (evt.lengthComputable) {
-      set(file, 'loaded', evt.loaded);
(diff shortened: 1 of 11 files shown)


tim-evans:
This is possible currently with `uploadBinary`

tim-evans:
I'm not opposed to this, just want to keep the API minimal / explicit. (It does look like a nice abstraction, but I'm not sure how often this would need to be overridden)

Subtletree:
`uploadBinary` is an option though it still forces the `Content-Type` which in my case is a signed header so s3 returns a `SignatureDoesNotMatch` error. It also forces a default `Accepts` header which needs to be whitelisted in CORS when it otherwise wouldn't need to be.

If it's that rare that more flexibility with the uploader is needed then I agree this is probably unnecessary. When rails 5.2 is released though I imagine there will be more people looking for this functionality.

Either way it was a good exercise putting it together

tim-evans:
Ahh, cool. Could you write your requirements here, or outline what's needed? I'm super unfamiliar with uploading binary files, so it's a feature that's more community driven.

tim-evans:
Actually, I think after sitting on this for a week, I'd be okay with this.

- [ ] Please write some tests for this ❤️ 
- [ ] Could you try to write to the existing API that exists (`uploadBinary` / `upload`)? Maybe these could be overridden using your solution- I don't really want to bump another major version for people because of a breaking change on a new API.
- [ ] Docs ✍️

Subtletree:
👍  will get onto it this week

RuslanZavacky:
@Subtletree this one is quite amazing, I think some issues that I have with uploading to S3, might be solved by this change. Do you think you can give some examples here, how to use it in the proper way (docs)? :)

Thank you!

tim-evans:
I would hold off until this gets merged for docs, since things may change.

Subtletree:
I've allocated some time this weekend to hopefully get this sorted.

@RuslanZavacky If you want help in the meantime feel free to ping me on the ember slack 🙂

tim-evans:
Sorry, not the intention to close this.

tim-evans:
Let me reopen the master branch temporarily 🙃

Subtletree:
I've hit a bit of a road block on my last couple of attempts to get this across the line.

 A `file` instance needs to `lookup` the uploader using the `owner`/`container`. The problem is that the `file` isn't registered so has no owner and my attempts at initializing it and adding it to the registry have failed so far.

The current code works by passing the `owner` down from the `fileService` to a `queue` instance to a `file` instance. It works fine but isn't ideal and would mean that if users of this addon wanted to manually create a `file` they would need to remember to pass a `owner`/`container` too.

I've tried several methods of injecting the `file` class into the registry in an initializer but haven't got it working so far.

I'll ask on slack if anyone with more experience in this area can take a look.

tim-evans:
Ahh, yeah. That'd be tricky. I can try to instantiate the file using the container; that may help quite a bit

tim-evans:
Sadly, I think it may break some of the niceness that comes from `File.fromDataURL`, etc.

Alonski:
Hey @Subtletree! I have taken over maintaining this addon.
Do you feel as if this PR is still relevant? What is needed to get this over the finish line?

Subtletree:
@Alonski Now I'm back from holiday I would like to revisit this. I think that the same thing could be accomplished much more simply by making the contentType customizable for `uploadBinary`, something like:

```
uploadBinary(url, opts) {
  // opts.contentType = 'application/octet-stream'; <-- old line
  opts.contentType = opts.contentType || 'application/octet-stream';

  return upload(this, url, opts, (request) => {
    return request.send(get(this, 'blob'));
  });
},
```

I think the 'uploader' idea would be useful in some situations but is mostly unnecessary. I'll have a better think about this later, whether to continue this pr or make a new one to edit `uploadBinary`

Good effort taking over the addon btw 🙂

Alonski:
@Subtletree That sounds like a good idea. Test it out and if it works we can get the new PR merged in :)

Alonski:
@Subtletree Would love to see this revisited. If not then this will be closed soon 🙂

Subtletree:
I'm closing this PR and am putting aside some time to work on the simpler method mentioned above.

I'm still using this branch in production.. so I'm keen to add this functionality into master and upgrade from v2.1.1 😅

PR #60, opened 2017-08-29
file.upload returns the promise created in http-request.js

### LINKED ISSUE #59: How to abort a file upload
There seems to be code in file and http-request to handle aborting an upload but I could not find the right way to do so.

If this is doable with the current code base, then doc is probably missing.
If it is not doable, then it might be useful adding it.

bartocc:
Previously, it was returning a new promise that could not be cancelled for example

fixes #59

### DIFF
--- a/addon/file.js
+++ b/addon/file.js
@@ -93,7 +93,9 @@ function upload(file, url, opts, uploadFn) {
   // Increment for Ember.Test
   inflightRequests++;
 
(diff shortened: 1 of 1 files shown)


bartocc:
I think I should add a test for this before the PR can be merged

tim-evans:
That would be lovely ❤️

Issue #31, opened 2017-04-19
Assertion Failed: Cannot call get with 'dataTransfer' on an undefined object.


EnotPoloskun:
This issue happens when I have my file browser(in my case - mac os Finder) over drag and drop area(dropzone). 

First upload goes through, but then, if I try to drag second file, this error happens. But if I drag second file out of dropzone and enter it again, upload will work. I found out, that `didEnterDropzone` doesn't happen in this case, which causes fail for this line `set(this[DATA_TRANSFER], 'dataTransfer', evt.dataTransfer);` in `didDrop`. So it means, that initialization of `this[DATA_TRANSFER]` doesn't happen, since `didEnterDropzone` is not triggered, and I have to drag file for second upload out of dropzone and then enter dropzone again. Everything works fine with first file upload.

If I add this line to `DragListener` `drop` function, then everything works fine.

```
      this._listener = null;
```

But I am not sure if it's not breaking something.

jelhan:
Should be fixed by #32 and #63.

PR #63, opened 2017-09-26
FF does not seem to have `dataTransfer` on dragleave

topaxi:
Tested in Nightly and FF55

Might be related to #31?

### DIFF
--- a/addon/components/file-dropzone/component.js
+++ b/addon/components/file-dropzone/component.js
@@ -147,7 +147,9 @@ export default Ember.Component.extend({
   didLeaveDropzone(evt) {
     set(this[DATA_TRANSFER], 'dataTransfer', evt.dataTransfer);
     if (this.isAllowed()) {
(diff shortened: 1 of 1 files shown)


addon/components/file-dropzone/component.js
```diff
@@ -147,7 +147,9 @@ export default Ember.Component.extend({
   didLeaveDropzone(evt) {
     set(this[DATA_TRANSFER], 'dataTransfer', evt.dataTransfer);
```
topaxi: I'm not sure if we should do this if `evt.dataTransfer` is null. Maybe skip this too if `evt.dataTransfer` is null?

tim-evans: Yeah, that may be 👍 

I will need to test this- could you provide a clear repro so I can get in and out of this quickly?

tim-evans:
This looks good; just tested. So weird.

PR #82, opened 2017-11-27
uploadPromise.then was lost as it was not returned

janmisek:


janmisek:
Actually it caused errors to be thrown twice. Because unassigned `uploadPromise.then` created orphan promise branch which was handled by global `RSVP.onerror` handler.

tim-evans:
Good catch! :/ Thanks for fixing this.

PR #89, opened 2017-12-01
Fix acceptance tests APIs

dwickern:
I made some fixes and improvements to the acceptance test helpers. While these are breaking changes (to the test helpers, not the app), they currently didn't work as documented anyway (if at all).

Changes to the mirage `upload` method:
- the callback now receives the original request e.g. so you can get the path arguments and query params
- the "file" is passed as the third argument to the callback instead. It's a `{ name, size, url }` hash as described in the readme (instead of `request.requestBody.file`)
- it fails with a 408 in case of timeout (I think this was the original intent?)

Changes to the `upload` async test helper:
- it's now a named export (previously the readme used both named and default import)
- it accepts a selector for the `<a>` element which the user created. Previously it took a selector for the hidden `<input type="file">` element which is awkward because the user doesn't even create that.
- it uses the ES6 File API directly. I extracted a `dataUrlToBlob` method for convenience when dealing with binary data.
- the helper no longer mutates `file.name`; this failed because it's a read-only property. Instead the user can pass the filename as the second argument to the `File` constructor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,24 +135,25 @@ In addition to the file list, there are properties that indicate how many bytes 
 import { upload } from 'ember-file-upload/mirage';
 
 export default function () {
-  this.post('/photos/new', upload(function (db, file) {
+  this.post('/photos/new', upload(function ({ db }, request, file) {
     let { name: filename, size: filesize, url } = file;
-    let photo = db.create('photo', { filename, filesize, url, uploadedAt: new Date() });
-    return photo;
+    return db.create('photo', { filename, filesize, url, uploadedAt: new Date() });
   }));
 }
 ```
 
 ```javascript
 import { upload } from '../../helpers/upload';
-import File from 'ember-file-upload/file';
+import { dataUrlToBlob } from 'ember-file-upload/file';
 
 moduleForAcceptance('/photos');
 
 test('uploading an image', async function (assert) {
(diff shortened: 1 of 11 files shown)


tim-evans:
Lemme take a look at this! Thanks for the PR

tim-evans:
Ugh. I had a review in progress, but left it. Sorry about leaving this hanging

ondrejsevcik:
It would be also great to move `test-support/helpers/upload.js` to `addon-test-support/helpers/upload.js` so this helper can be easily imported with `import upload "ember-file-upload/test-support/upload.js` becaus that doesn't work at the moment.

tim-evans:
@ondrejsevcik I think that'll need to be symlinked or copied so we maintain backwards compatibility, but yes! That sounds like a great idea

ondrejsevcik:
I think that re-exporting it should do the thing.

dwickern:
I don't intend to work on this, so if someone else wants to pick it up go ahead.

Here's how I currently test uploads:
```js
let input = find('input[type=file]')[0];
let file = new File([ new Blob(['hello world']) ], 'file.txt', {type: 'text/plain'});
Object.defineProperty(input, 'files', { value: [ file ]});
await triggerEvent(input, 'change');
```
This works in Chrome, FF, and with some basic `File` and `FormData` polyfills, IE11.

PR #101, opened 2018-01-25
Add support for uploading Files from Data URL

seanpdoyle:
When a File generated by `File.fromDataURL` is uploaded in test, the
guard clause in the Queue service (i.e. `if (fileBlob instanceof Blob)`)
rejects the `fileBlob`.

In Chrome, a File generated by `File.fromDataURL` will have the `Blob`
available under the `.blob` property.

The `upload` test helper wasn't declared with
`Ember.Test.registerAsyncHelper`, so it wasn't waiting on or blocking
other asynchrouns helpers.

Once declared with `registerAsyncHelper`, we'll want to import it once
and refer to it as if it were `click` or `visit`.

This commit adds documentation to the README about adding an `import`
statement to applications' test helpers.

The `import` statements in the example code were replaced by `/* global
upload */` magic comments. Ideally, applications will add an `upload` to
their linter configuration files as a globally accessible function.

Add dummy application acceptance tests
---

Introduce tests that exercise using the plugin _in a consumer
application_.


tim-evans:
Probably poor documentation on my part, but https://github.com/tim-evans/ember-file-upload/blob/latest/addon/file.js#L347 should work

tim-evans:
Perhaps I was wrong, sorry about that! Thanks for the PR, I'll take a peek at this

seanpdoyle:
Do you know what part of this diff broke CI?

README.md
```diff
@@ -144,15 +151,15 @@ export default function () {
 ```
 
 ```javascript
-import { upload } from '../../helpers/upload';
+/* global upload */
```
seanpdoyle: I'm a bit behind on my own projects' `ember-cli` version, so I don't know how to go about either `import` this helper (which we're now declaring with `Ember.Test.registerAsyncHelper`) or configuring our linter to consider it globally accessible.

Addons used to modify the `"global"` key in projects' `.jshintrc` and `tests/.jshintrc` files, but now we use `.eslintrc`.

What do you suggest?

README.md
```diff
@@ -162,14 +169,14 @@ test('uploading an image', async function (assert) {
...
-import upload from '../helpers/upload';
+/* global upload */
 
 moduleForAcceptance('/notes');
 
 test('showing a note', async function (assert) {
   let file = File.fromDataUrl('data:text/plain;base64,SSBjYW4gZmVlbCB0aGUgbW9uZXkgbGVhdmluZyBteSBib2R5');
 
-  await upload('#upload-note', file, 'douglas_coupland.txt');
+  await upload('#upload-note input[type="file"]', file, 'douglas_coupland.txt');
```
seanpdoyle: Other examples with templates use `id` selectors (like `"#upload-image"`) to reference elements that aren't the `<input>`

Since these examples don't have corresponding example templates, I've added the `input[type="file"]` selector here to indicate that the selector we want is for the file upload _input_.

test-support/helpers/ember-file-upload.js
```diff
@@ -0,0 +1,5 @@
+import upload from './upload';
+
+export {
```
seanpdoyle: This export might be unnecessary, since we're importing `./upload` for side-effects only.

test-support/helpers/upload.js
```diff
@@ -1,19 +1,36 @@
...
+export default function(selector, file, filename) {
+  const scope = window || global;
+  const fakeApplicationFromScope = {
+    testHelpers: {
+      findWithAssert: scope.findWithAssert,
+      triggerEvent: scope.triggerEvent,
+    }
+  };
+
+  return upload(fakeApplicationFromScope, selector, file, filename);
```
seanpdoyle: This exported function is an _attempt_ to maintain backward compatibility with client applications that are `import`-ing the `upload` helper.

I'm not sure if this is a good pattern to adopt, or even if it'll work, but the idea is that since `triggerEvent` and `findWithAssert` were previously invoked from the global scope, we should pass those in so that they'll continue to be called that way.

What do you think?

seanpdoyle:
I'm not sure what's causing CI to fail for this project. Do you?

On the application that motivated this PR, we had issues running the test helpers in Phantom 2.1, but had no problems with Chrome.

README.md
tim-evans: I definitely need to go and update this to Ember 2.18, since it's the last LTS before 3.0

tim-evans: I think this is alright how you did it

README.md
tim-evans: That works! I think I chose not to so we didn't expose the private internals of the `{{file-upload}}` template.

test-support/helpers/ember-file-upload.js
tim-evans: Sounds right- I think that I didn't do this because Ember's testing story is moving away from globals. But since I have no idea where that's in the pipeline of things, it makes sense to use what's provided out-of-the box.

test-support/helpers/upload.js
tim-evans: I think that's a good idea here- I would prefer to have something else cause a major version release, but if it's necessary, it's necessary.

tim-evans:
Thanks for the writeup- I think I have some cycles this week to look through this.

As for testing, ugh. Looks like there's something wonky with Travis. I think I restarted it in hopes that it would resolve the issue on its own (which it didn't look like it did). I'm going to spend some time upgrading the addon to 2.18 and then revisit these issues that you're encountering. Hopefully that'll resolve these problems 🤞

tim-evans:
Ok; just pushed changes for Ember 2.18- hopefully that'll help with some of this

tim-evans:
(I handled the merge conflict here)

Alonski:
@seanpdoyle Hey! I have taken up maintaining this addon. Could you please rebase/merge so that this PR can be hopefully merged in soon?

Alonski:
@seanpdoyle Would love to see this revisited. If not then this will be closed soon 🙂

jelhan:
Closing as it seems to be stale. Feel free to reopen or open a new pull request as soon as the changes have been rebased on latest master.

PR #111, opened 2018-03-01
Support for files without extension. e.g: "FILE"

janmisek:
Previous version was failing on exception as regex match returned null


tim-evans:
This is great, thanks!

PR #121, opened 2018-03-22
Copy item details out of DataTransfer object when scheduling a dragenter handler, so they can be reliably accessed

### LINKED ISSUE #104: dropzone.valid is not working in Chrome
Hi there, 

I'm using the latest Chrome for Mac (64.0.3282.140) and I just noticed that the `dropzone.valid` equals true even if the file type doesn't match the accept-pattern. I tried it on your demo site to double check it. A file called `test` or `test.txt` succeeds even though the pattern is `accept="image/*,video/*,audio/*"`. Works fine in Firefox (58.0.1)

lukemelia:
This is necessary because in Chrome, by design, the contents of evt.dataTransfer.items are only available during the event frame that the drag event is being handled. [bug-tracker-link]

Fixes #104

[bug-tracker-link]: https://bugs.chromium.org/p/chromium/issues/detail?id=137231#c1

### DIFF
--- a/addon/components/file-dropzone/component.js
+++ b/addon/components/file-dropzone/component.js
@@ -130,7 +130,8 @@ export default Component.extend({
     let dataTransfer = DataTransfer.create({
       queue: get(this, 'queue'),
       source: evt.source,
-      dataTransfer: evt.dataTransfer
+      dataTransfer: evt.dataTransfer,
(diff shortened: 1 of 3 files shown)


tim-evans:
Thanks @lukemelia ❤️

tim-evans:
Released in 2.3.0

PR #125, opened 2018-03-28
Don't depend on jQuery

raido:
By removing jQuery usage, enables to use this addon in Ember applications where jquery.js is set to null in ember-cli-build.js.

### DIFF
--- a/addon/components/file-upload/component.js
+++ b/addon/components/file-upload/component.js
@@ -1,6 +1,7 @@
 import { assert } from '@ember/debug';
 import Component from '@ember/component';
 import { inject as service } from '@ember/service';
+import { DEBUG } from '@glimmer/env';
 import {
   getProperties,
   setProperties,
@@ -9,13 +10,6 @@ import {
 } from '@ember/object';
 import layout from './template';
 import uuid from '../../system/uuid';
-
-const VALID_TAGS = ['a', 'abbr', 'area', 'audio', 'b', 'bdo', 'br', 'canvas', 'cite',
-                    'code', 'command', 'datalist', 'del', 'dfn', 'em', 'embed', 'i',
-                    'iframe', 'img', 'kbd', 'mark', 'math', 'noscript', 'object', 'q',
-                    'ruby', 'samp', 'script', 'small', 'span', 'strong', 'sub', 'sup',
-                    'svg', 'time', 'var', 'video', 'wbr',
-                    'path', 'g', 'use', 'circle'];
 
 /**
   `{{file-upload}}` is an element that will open a dialog for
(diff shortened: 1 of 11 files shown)


raido:
It is known that parseFromString method is not available in older browsers, so maybe polyfill should be included?

Or maybe the responsibility of parsing the responseText should be left up the application to decide, simply return "responseText" and that's it?

addon/components/file-upload/component.js
```diff
@@ -164,7 +164,7 @@ export default Component.extend({
   didInsertElement() {
     let id = get(this, 'for');
     assert(`Changing the tagName of {{file-upload}} to "${get(this, 'tagName')}" will break interactions.`, get(this, 'tagName') === 'label');
-    this.$('*').each(function (_, element) {
+    this.element.querySelector('*').each(function (_, element) {
```
tim-evans: This probably needs to be `querySelectorAll`.

tests/unit/system/http-request-test.js
```diff
@@ -85,7 +85,7 @@ module('http-request', function(hooks) {
       assert.deepEqual(response.headers, {
         'Content-Type': 'text/html'
       });
-      assert.equal(response.body[0].textContent, 'ok');
+      assert.equal(response.body.body.textContent, 'ok');
```
tim-evans: This probably shouldn't change with the changes to remove jQuery

tim-evans:
This addon has semver and I try veeery hard not to release major versions, so we'll need to handle that.

What browsers don't support this?

raido:
DOMParser.parseFromString is not supported on IE < 10 and Android 4.4 for example - https://caniuse.com/#search=parse

Currently DOMParser is used only for XML parsing, for HTML i switched to "document.implementation.createHTMLDocument" API.

tim-evans:
Ember doesn't support IE < 10, so let's drop support here for those browsers. :)

raido:
Well, Ember 2.x users still might need it. Which polyfills you mean? Those utils i introduced are native implementations.

tim-evans:
Fair point. I would prefer not to include it, but if there are specific folks that are supporting these browsers, then there's a possibility for this to be included.

tim-evans:
I'm currently reviewing; was a bit bogged down with work yesterday.

tim-evans:
Ahh, I believe jQuery is being used in the test helpers. If the test-helper codemod is run, this error *should* go away.

raido:
It failed because older versions of Ember use jQuery to dispatch events, simple fix was to install - https://github.com/rwjblue/ember-native-dom-event-dispatcher

tim-evans:
Ok! I'm going to merge this, and bump this as a minor version. There *should* be no breaking changes here 🤞.

Would it be alright if I prodded you if any issues come up from this release @raido ?

PR #126, opened 2018-04-02
ember-native-dom-event-dispatcher should be in ember-try config instead

raido:
It simply makes way more explicit why ember-native-dom-event-dispatcher is needed in the first place and removes the warning that this addon should not be included in Ember 3.0 apps.


tim-evans:
Thanks!

PR #137, opened 2018-04-18
🔒 eliminate inline style

jamesarosen:
Previously, `{{file-upload}}` would render an `<input>` with `style="display: none;"`.

A good [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) will block inline styles.

This commit replaces `style="display: none;"` with `hidden`. The `[hidden]` attribute is an HTML5 standard. Browser support is [great](https://caniuse.com/#feat=hidden) on modern browsers.

Unfortunately, a rule as simple as
```css
input { display: block; }
```
will override the browser's default handling for `[hidden]`. (See [this jsfiddle](https://jsfiddle.net/j74ooj4b/).) For this reason, I added an `addon/styles/addon.css` with a powerful selector and rule to ensure the input stays hidden.

### DIFF
--- a/addon/styles/addon.css
+++ b/addon/styles/addon.css
@@ -1,0 +1,3 @@
+.file-upload input[hidden] {
+  display: none !important;
+}
--- a/addon/system/uuid.js
+++ b/addon/system/uuid.js
@@ -26,8 +26,8 @@ let random = (function () {
       }
 
       numbers[i] = r >>> ((i & 0x03) << 3) & 0xFF;
-      return numbers;
     }
+    return numbers;
   };
 })();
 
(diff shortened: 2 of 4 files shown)


addon/styles/addon.css
```diff
@@ -0,0 +1,3 @@
+.file-upload input[hidden] {
+  display: none !important;
+}
```
jamesarosen: ember-cli automatically appends this file to `vendor.css`, so we don't need to change any configuration.

addon/system/uuid.js
```diff
@@ -26,8 +26,8 @@ let random = (function () {
       }
 
       numbers[i] = r >>> ((i & 0x03) << 3) & 0xFF;
-      return numbers;
     }
+    return numbers;
```
jamesarosen: This was causing the build to fail. It's not related to the work at-hand, so if you want, I can put it in a separate PR.

tim-evans: This is fine here :)

tim-evans:
I’ll get this out tomorrow ~ :sparkles:

Issue #138, opened 2018-04-19
Replace jQuery's findWithAssert in test helper upload


imalexsq:
helpers/upload.js is using jQuery's findWithAssert which was removed from Ember 3.0. Ember test gives 'findWithAssert is not defined' error.

imalexsq:
I've been trying to solve this issue and here are my findings so far:

1. `let input = findWithAssert(selector)[0];` can be replaced with Ember's find method like so `fileInput = find('.file-upload input');`

2. `file.name = filename;` does not work as file.name is read-only string

3. After a file is added, I cannot seem to access it through FileList object in the console. I'm not sure where this file is stored.

tim-evans:
See https://github.com/tim-evans/ember-file-upload/pull/89

If you could help kick that PR over the finish line, it would be very appreciated ♥️

imalexsq:
@tim-evans I'd love to contribute, but I'll need some help. Can you chat over EmberJS Slack channel?

tim-evans:
That i can! I’m @tce over there. I’ll probably respond tomorrow during NYC work hours

Issue #70, opened 2017-10-10
Blueprints don't install test-support/helpers/upload.js into the project


psbanka:
The documentation for the acceptance-test helpers indicates that you should be able to import `upload` from `../../helpers/upload`, but that file is not installed. I don't know if it was intended that this be installed via a blueprint or not? 

doing this manually with `cp node_modules/ember-file-upload/test-support/helpers/upload.js tests/helpers/` seemed to do the trick, however.

tim-evans:
Oh, that's some weird thing that Ember does, merging the test helper in test mode. Lemme get some docs for you :)

tim-evans:
For some info, see here: https://williamsbdev.com/posts/Ember-CLI-test-support/

`test-support` should be merged in by ember-cli. Something may have changed in the most recent of ember. :/

tim-evans:
Thanks for the report

jelhan:
Outdated. It's exported from 'ember-file-upload/test-support' since quite some time.

PR #146, opened 2018-05-08
Fix upload test-helper

### LINKED ISSUE #144: upload test-helper does not trigger upload
Hi, I can't figure out what am I doing wrong, but the upload helper is not working for me. I have a following acceptance test

```js
module('Acceptance | upload', function(hooks) {
  setupApplicationTest(hooks);

  test('upload works', async function(assert) {
    await visit('/');

    let file = File.fromDataURL('data:image/gif;base64,R0lGODdhCgAKAIAAAAEBAf///ywAAAAACgAKAAACEoyPBhp7vlySqVVFL8oWg89VBQA7');
    await upload('#upload-photo', file, 'smile.gif');

    let photo = server.db.photos[0];
    assert.equal(photo.filename, 'smile.gif');
  });
});
```

which is basically what's in the README. I can see that upload helper is being called with my input selector, but when I put a `debugger;` into the JS code, I can also see that my `onaddfile` handler is not called and thus nothing is uploaded to the server.

What am I doing wrong?

EDIT:

After a bit of investigation, I've found this line https://github.com/tim-evans/ember-file-upload/blob/latest/addon/queue.js#L68. The problem here is that in the example we use File (ember model) but this code requires File (native browser object).

I also had to comment out this line https://github.com/tim-evans/ember-file-upload/blob/latest/test-support/helpers/upload.js#L6 because File.name is readonly. 

Now it's working. 

I can do a PR to fix this issue, if you're interested.

ondrejsevcik:
This should fix #144, #138 and #70 
Inspired by #89


test-support/helpers/upload.js
```diff
@@ -1,19 +0,0 @@
-/*global triggerEvent, find, findWithAssert */
```
ondrejsevcik: I've moved this one into `addon-test-helpers` to allow for easy import

`import {upload} 'ember-file-upload/test-support'`

We probably don't need to keep backward compatibility in here as didn't work anyway.

tim-evans: I think this worked for prior versions of ember (pre 3.0)

tim-evans: Could you keep this, just in case? I know it may be broken, but I'd prefer not to major version bump.

ondrejsevcik: It's back.

tim-evans:
:heart: I’ll get this in and released soon!

PR #154, opened 2018-06-15
Make `name` property clearer in docs

geekygrappler:
As suggested in https://github.com/tim-evans/ember-file-upload/issues/151


tim-evans:
Thanks! It looks like the docs website could use some more love with regard to this, but it's there, so this is a good step! ❤️

PR #160, opened 2018-07-12
enable dropping directories

bekzod:
chrome and some other browsers support [webkitGetAsEntry](https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/webkitGetAsEntry) which enables dropping directories to dropzone

### DIFF
--- a/addon/components/file-dropzone/component.js
+++ b/addon/components/file-dropzone/component.js
@@ -19,6 +19,45 @@ let supported = (function () {
 
 const dragListener = new DragListener();
 
+function walkDir(entry, path = '', ret = []) {
+  if (entry.isFile) {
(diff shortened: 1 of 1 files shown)


tim-evans:
I'm interested in this, but am worried about having this feature in that may get dropped at any time by browser vendors (since it's non-standard)

Alonski:
@bekzod Would love to see this revisited. If not then this will be closed soon 🙂

jelhan:
I don't think the `webkitGetAsEntry` API is a safe bet for the future. No vendor seems to support it. Google Chrome team pushes [Native File System API](https://wicg.github.io/native-file-system/).

jelhan:
Closing as this seems to be stale. Feel free to repoen or open another pull request based on this one.

PR #161, opened 2018-07-14
Tagless file-upload component

jonnii:
I'm running into a problem where the pre-defined markup for the file selector isn't compatible with the design system markup I need to generate (https://www.lightningdesignsystem.com/components/file-selector/#site-main-content). Specifically I need to do something like:

```
<input type="file"  />
<label class="slds-file-selector__body" for="file-upload-input-01" id="file-selector-secondary-label">
    ...
</label>
```

While I can get the component to work (it's great by the way!) the selector the design system uses is very (overly?) specific when it comes to styling in the `disabled` state.

What I'd like to do is make the `file-upload` component tag-less and then use presentational components to allow the markup to be customized, you'll end up being able to do something like:

```
{{#file-upload as |upload queue|}}
  <label for={{upload.id}} />
  {{upload.input}}
{{/file-upload}}
```

What are your thoughts? Have I overlooked something?

p.s. don't focus too much on the code here - I was just seeing if this would even be possible and needed something to allow me to open the PR.


tim-evans:
Hi @jonnii 

So, ugh. Let's try to figure out the best way to do this without removing the validation stuff- I know people are pretty unhappy about it, but the amount of time it saves in the end is a lot, and it improves accessibility.  I *think* there's some things that we can do to make this easier to manage. I would definitely be interested in using a tagless component, but am a bit concerned about changing the structure of this and breaking everyone's CSS.

It looks like the design system wants to get:

```hbs
<label>{{yield}}</label>
<input>
```

Instead of:

```hbs
<label>
  <input>
</label>
```

The latter is leveraging the system to trigger the file upload dialog w/o resorting to an invisible div floating around the screen. The former is totally possible as well, but maaaay end up be confusing, since then we'd have 2 labels that refer to the same element. That sounds ok to me at first glance.

I'm willing to look into changing the markup, but don't want to remove the validation. That's there to prevent your dom from being rendered incorrectly. Browsers occasionally will move the dom element outside of the label bc it's invalid 😭  Again, this isn't really something I remember recently, but it's definitely been something that I've wasted several days worth of time debugging and having this log here prevents a whole load of other people from making similar mistakes (also, not messing with accessibility).

jonnii:
The validation stuff you're referring to is the stuff in the `didInsertElement` hook correct? If so I only took that out temporarily - I can see why that would be a useful check! 

> I would definitely be interested in using a tagless component, but am a bit concerned about changing the structure of this and breaking everyone's CSS.

The challenge here is if someone has set a `class` on the file-upload component, this wouldn't be propagated to the yielded template (we can do this explicitly though to maintain backwards compatibility) or do something more drastic like having another component which is tagless and perhaps deprecating the existing one.

jonnii:
@tim-evans what do you think of the approach here? I've relaxed the check to include `label` and am no longer skipping elements which have the `for` that matches. I've unfortunately had to use an intimate API call to get the parent element for the tag-less component, but this is a very commonly used method and won't be removed without an RFC.

tim-evans:
Ooh, nice! This looks like the markup is identical to what it was previously- did you want to alter that at all, or is as-is good, and you’ll override the component template?

jonnii:
There's more I want to do here before it's ready to merge, but I think this is a step in the right direction. As it stands this is a breaking change and if we're in here breaking things, what else do you want to change/fix?

tim-evans:
Nothing comes to mind! Thanks for tackling this ❤️❤️❤️

Alonski:
@jonnii Hey man! I have taken up maintaining this addon. Do you feel as if this PR could use some more work, and we can get it merged in eventually?

jonnii:
@Alonski it needs a rebase, docs and some clean up but I think it's good enough as an incremental step. Let me see if I can find some time this evening to whip it into shape.

Alonski:
Thanks @jonnii! Whenever you have time would be awesome

jonnii:
@Alonski what do you think we should do about the breaking change?

> The challenge here is if someone has set a class on the file-upload component, this wouldn't be propagated to the yielded template (we can do this explicitly though to maintain backwards compatibility) or do something more drastic like having another component which is tagless and perhaps deprecating the existing one.

This is one of the challenges from moving to this style of component.

Alonski:
@jonnii I think creating two components and depreciating the old one sounds good. It is something Ember developers know pretty well. We could also have a feature flag that turns it on or off. This way we can only have one component and not two.

Alonski:
@jonnii Would love to see this revisited. If not then this will be closed soon 🙂

jonnii:
we can close this, i'm not going to have the time to get it finished =(

Alonski:
Ok @jonnii Thanks!
If you ever want to pick this up again you know where to find this :)

PR #167, opened 2018-08-31
Allow setting capture device for file input

### LINKED ISSUE #130: Camera capture attribute
Hi, we are using this plugin for mobile app and don't want to use Cordova to capture image. will it possible to add attribute to capture camera? It looks like to just add an attribute called capture pass same to template default will be nothing. 

Please let me know what do you think about it? if yes I'll send you pull request for same

Thanks

lougreenwood:
Fixes #130

### DIFF
--- a/addon/components/file-upload/component.js
+++ b/addon/components/file-upload/component.js
@@ -104,7 +104,8 @@ const component = Component.extend({
   accept: null,
 
   /**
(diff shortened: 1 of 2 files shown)


tim-evans:
Brilliant; thanks @lougreenwood ❤️

PR #169, opened 2018-09-04
Fix testing example

simonihmig:
Update testing example to fix the correct signature of the mirage `upload` helper's callback.

The callback receives the `request`, not directly the `file` here: https://github.com/tim-evans/ember-file-upload/blob/latest/addon/mirage/index.js#L56

Great addon btw! 👍


tim-evans:
Oh, thanks!

PR #178, opened 2018-10-08
Fix for - Dropzone triggering events out of scope->which should not

aysha-rafa:
Regarding this issue - https://github.com/tim-evans/ember-file-upload/issues/177

The listeners were added to body, changed it to the selectors so that it is triggered only when it is on the target.

### DIFF
--- a/addon/system/drag-listener.js
+++ b/addon/system/drag-listener.js
@@ -15,7 +15,7 @@ export default class {
     this._events = A();
   }
 
-  beginListening() {
+  beginListening(selector) {
     let handlers = this._handlers = {
       dragenter: bind(this, 'dragenter'),
       dragleave: bind(this, 'dragleave'),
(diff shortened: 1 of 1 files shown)


addon/system/drag-listener.js
```diff
@@ -15,15 +15,15 @@ export default class {
     this._events = A();
   }
 
-  beginListening() {
+  beginListening(selector) {
```
tim-evans: This is intentional, because there's issues where drag listeners don't get called consistently in browsers 😭 (The events will sometimes skip elements or repeat events on elements)

I _think_ the bug is with the `findListener` function.

tim-evans: It seems that `evt.target.find(target)` should be removed in the `findListener` function. That will return `true` for all cases.

aysha-rafa: Thanks very much for your reply ❤️ I have created a new pull request!

PR #182, opened 2018-10-09
Adds support for additional SVG tags

jakemauer:
Adds additional SVG tags to the VALID_TAGS constant:

• circle
• clipPath
• line
• mask
• radialGradient
• rect
• polygon
• polyline


tim-evans:
@jakemauer fabulous!

Issue #197, opened 2018-12-05
'aborted' uploads can not be uploaded again


toovy:
An upload can be aborted by the client like described in https://github.com/tim-evans/ember-file-upload/issues/59.

Currently it is not possible to upload a file again if the client aborted it via uploadPromise.cancel().

Simply adding `aborted` to the array in L54 solves the issue, the file can be uploaded again if the user decides to do so:

https://github.com/tim-evans/ember-file-upload/blob/5a1becfd9363a2d16a09c8b6fe468d64741e7881/addon/file.js#L54

Is there any technical background why this should not be possible? I could create a PR.

jelhan:
Should be fixed by #200.

PR #200, opened 2018-12-10
files that were cancelled can be uploaded again

toovy:
Fix for https://github.com/adopted-ember-addons/ember-file-upload/issues/197


toovy:
@Alonski that makes sense and honestly I wanted to, but as there is no case testing the other states as well this is not a simple task, it would require much more work than I can spare at the moment.

PR #204, opened 2019-01-24
Added a possibility to run tests that use 'upload' test-helper in Microsoft browsers

kazimirchuk:
Hi there! I hope it's okay that I created a PR right away without creating an issue (I scanned the existing ones before doing so) 🙏

The problem I stumbled upon:
* I run acceptance tests on my app in different browsers. It turned out that Microsoft ones (like Edge and IE) don't support `File` API. Proof link - https://developer.mozilla.org/en-US/docs/Web/API/File/File 😢
* The only way to have tests passing in Microsoft browsers while still being written in the same fashion as this documentation suggests is to use `Blob` instead of the `File`.
* Helper itself can work both with `File` and `Blob` instances without any problems, however before doing any work it checks instanceof File, which in my case basically translates to *This helper can be ran only in non-Microsoft browsers*

The solution is to allow both instances of `File` and `Blob`.

I don't know whether `Blob` is not allowed for a reason or not, but anyway I hope it's a useful change. 🤷‍♂️

Thanks 🙇

### DIFF
--- a/addon-test-support/index.js
+++ b/addon-test-support/index.js
@@ -4,7 +4,7 @@ import { assert } from '@ember/debug';
 export async function upload(selector, file) {
   let input = find(selector);
   assert(`Selector '${selector}' is not input element.`, input.tagName === 'INPUT');
-  assert(`File must be instance of File type`, file instanceof File);
+  assert(`File must be instance of File type`, file instanceof File || file instanceof Blob);
 
   // This hack is here since we can't mock out the
   // FileList API easily; we're taking advantage
--- a/addon-test-support/index.js
+++ b/addon-test-support/index.js
@@ -4,7 +4,7 @@ import { assert } from '@ember/debug';
 export async function upload(selector, file) {
   let input = find(selector);
   assert(`Selector '${selector}' is not input element.`, input.tagName === 'INPUT');
-  assert(`File must be instance of File type`, file instanceof File || file instanceof Blob);
+  assert(`File must be instance of File/Blob type`, file instanceof Blob);
 
   // This hack is here since we can't mock out the
   // FileList API easily; we're taking advantage
--- a/tests/acceptance/upload-test.js
+++ b/tests/acceptance/upload-test.js
@@ -8,7 +8,7 @@ module('Acceptance | upload', function(hooks) {
   setupApplicationTest(hooks);
   setupMirage(hooks);
 
-  test('upload works', async function(assert) {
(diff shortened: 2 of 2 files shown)


Alonski:
Thanks @kazimirchuk ! And thanks @knownasilya for the ping. I'll have time to take a look at this over the weekend and once it's merged I'll push it a new release

addon-test-support/index.js
```diff
@@ -4,7 +4,7 @@ import { assert } from '@ember/debug';
 export async function upload(selector, file) {
   let input = find(selector);
   assert(`Selector '${selector}' is not input element.`, input.tagName === 'INPUT');
-  assert(`File must be instance of File type`, file instanceof File);
+  assert(`File must be instance of File type`, file instanceof File || file instanceof Blob);
```
Alonski: Maybe change the assertion to mention `Blob`?

`File must be instance of File or Blog type`?

Alonski: You could also just check if `file instanceof Blob` since File is based on Blog:
https://developer.mozilla.org/en-US/docs/Web/API/File

Alonski:
One thing to note is that Blobs don't have a `name` property. 
https://developer.mozilla.org/en-US/docs/Web/API/Blob#Properties
However for the current implementation this is needed. I wonder what should happen if a user passes in a Blob without a `name`. 

What do you think @kazimirchuk ?

Other than that this looks good,

kazimirchuk:
@Alonski well, if we don't set any `name` property explicitly to the Blob object it anyway sends the file nicely because we declare a name for an upload input component itself in the template like so:
```
{{#file-upload name="avatar" ... }}{{/file-upload}}
```

So from what I understand: 
* Since input anyway gets populated with Blobs without a file name defined and gets them sent to a backend, then it means addon doesn't really care about file name itself? 🤷‍♂️
* Consumers of this addon that choose to use Blobs in a real app for some reason should take responsibility for figuring out if they rely on filename data or not and handling this nicely. 

To be honest, the idea behind this change was to give a slightly more flexible way of testing stuff out, and in this context it does seem okayish, but if we try to look at it from a standpoint of promoting the usage of Blobs with this addon then I'm not sure if it worth it 🤔. But yeah, to summarize it - I didn't find any crashes because of undefined filename

Alonski:
Thanks @kazimirchuk !
Now I just need to figure out how to release this 😅 
First time

Issue #206, opened 2019-01-31
ember-cli-update should be in devDependencies


andreyfel:
ember-cli-update in dependencies brings too many extra dependencies to users of the addon.

andreyfel:
All good with the release of 2.6.1! Thanks @Alonski!

PR #207, opened 2019-01-31
Move ember-cli-update to dev dependencies

andreyfel:
ember-cli-update shouldn't be a dependency of apps which use ember-file-upload.


andreyfel:
Addresses #206

PR #231, opened 2019-02-25
Fix MIME type checks

bendemboski:
Using a regular expression to check MIME types was problematic and unnecessary. Problematic because it wasn't escaping characters, so for example, `image/svg+xml` would not match because the `+` was interpreted as a 1-or-more modifier on the `g`. Unnecessary because the accept attribute only accepts a very limited set of clearly enumerated type wildcards (see https://html.spec.whatwg.org/multipage/input.html#file-upload-state-(type=file)).

So, match more tightly and correctly doing string comparisons, and add some unit tests for this (and to verify case-insensitivity, which was missing for mime types).

### DIFF
--- a/addon/system/data-transfer.js
+++ b/addon/system/data-transfer.js
@@ -82,10 +82,16 @@ export default EmberObject.extend({
       let extensions = A(tokens.filter(function (token) {
         return token.indexOf('.') === 0;
       }));
-      let mimeTypes = A(A(tokens.filter(function (token) {
+      let mimeTypeTests = A(A(tokens.filter(function (token) {
         return token.indexOf('.') !== 0;
       })).map(function (mimeType) {
-        return new RegExp(mimeType);
+        return function(type) {
+          if (A([ 'audio/*', 'video/*', 'image/*' ]).includes(mimeType)) {
+            return type.split('/')[0] === mimeType.split('/')[0];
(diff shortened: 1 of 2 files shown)


Alonski:
Hmm Ember Beta and Ember Canary are failing builds unrelated to this PR. If you could also help out with getting these tests passing that would be awesome. If not I will hopefully get to it this week or next week :)

bendemboski:
I think I should be able to get to that in later today.

Alonski:
Awesome @bendemboski thanks!

bendemboski:
Okay, investigated a bit, and the failure is because `ember-beta` and `ember-canary` broke `ember-concurrency` (which is used by `ember-addon-docs`). It sounds like the [PR](https://github.com/emberjs/ember.js/pull/17649) reverting the breaking changes will be released to beta today or tomorrow (see [here](https://github.com/machty/ember-concurrency/pull/277#issuecomment-467558695)), and when that happens, the build will magically start passing 😄 

When that gets released I'll poke this branch to get it rerun the build and ping you once it's ready and passing.

Alonski:
Awesome detective work! Let's wait for that and once tests are passing let's get this stuff merged!

bendemboski:
@Alonski build passing now!

PR #236, opened 2019-04-24
Fix computed-property.override deprecation

### LINKED ISSUE #235: Deprecation with Ember@3.9
After upgrading the app which is using ember-file-upload I'm getting this deprecation:
`computed-property.override`
`file-upload#for computed property was just overriden`
Need to fix it here: https://github.com/adopted-ember-addons/ember-file-upload/blob/05f734cb1e7126efb460bdbdfa49b8135511219d/addon/components/file-upload/component.js#L91
Probably we can use init() instead of computed property to fix that.

andreyfel:
Ember@3.9 deprecates overriding a computed property with .set().

Use didReceiveAttrs() hook to set the default value of the "for" property instead of using a computed property.

Fixes #235

### DIFF
--- a/addon/components/file-upload/component.js
+++ b/addon/components/file-upload/component.js
@@ -91,6 +91,9 @@ const component = Component.extend({
   for: computed({
     get() {
       return `file-input-${uuid.short()}`;
+    },
+    set(key, value) {
+      return value;
     }
   }),
 


addon/components/file-upload/component.js
```diff
@@ -88,11 +88,7 @@ const component = Component.extend({
 
   attributeBindings: ['for'],
 
-  for: computed({
-    get() {
-      return `file-input-${uuid.short()}`;
-    }
-  }),
+  for: undefined,
```
Alonski: After looking at this again. Would it not be possible to have a setter for this computed prop instead of setting it in didReceiveAttrs?
Would that work?

andreyfel: Sure, we can use that approach as well. Changed the code to use setter.

PR #274, opened 2019-09-28
request params are not passed through to mirage handler

jelhan:
If using the mirage upload handler provided by `ember-file-upload/mirage` the request params are not passed through. This blocks usage of upload handler together with dynamic segments on URLs.

This PR fixes that bug and maybe others caused by replacing the `request` object. Instead it only replaces the `requestBody` property of it. It also adds test coverage for passing through schema and `request.params` to prevent a regression.

### DIFF
--- a/addon/mirage/index.js
+++ b/addon/mirage/index.js
@@ -43,17 +43,15 @@ export function upload(fn, options={ network: null, timeout: null }) {
           });
 
           metadata.then((metadata) => {
-            let response = {
-              requestBody: Object.assign({
-                [file.key]: metadata
-              }, data)
-            };
+            request.requestBody = Object.assign({
+              [file.key]: metadata
+            }, data);
             if (timedOut) {
               resolve(new Response(408));
               return;
             }
 
-            resolve(fn.call(this, db, response));
+            resolve(fn.call(this, db, request));
           });
         } else {
           request.upload.onprogress({
--- a/tests/integration/components/mirage-handler-test.js
+++ b/tests/integration/components/mirage-handler-test.js
@@ -1,0 +1,42 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import hbs from 'htmlbars-inline-precompile';
+import { upload as uploadHandler } from 'ember-file-upload/mirage';
+import { upload } from 'ember-file-upload/test-support';
+import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
+
+module('Integration | Component | mirage-handler', function(hooks) {
+  setupRenderingTest(hooks);
+  setupMirage(hooks);
+
+  test('upload handler passed schema and request.params throw', async function(assert) {
+    let self = this;
+    this.server.post('/folder/:id/file', uploadHandler(function(schema, request) {
+      assert.step('mirage-handler');
+
+      assert.equal(schema, self.server.schema, 'schema is provided');
+      assert.ok(typeof request.params === 'object', 'params property on request is an object');
+      assert.equal(request.params.id, '1', 'value of dynamic segment is present on params object');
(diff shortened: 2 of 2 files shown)


addon/mirage/index.js
```diff
@@ -43,17 +43,15 @@ export function upload(fn, options={ network: null, timeout: null }) {
...
+            request.requestBody = Object.assign({
+              [file.key]: metadata
+            }, data);
             if (timedOut) {
               resolve(new Response(408));
               return;
             }
 
-            resolve(fn.call(this, db, response));
+            resolve(fn.call(this, db, request));
```
Alonski: Is it possible to do this without mutating the request object? I prefer a more functional approach with regards to immutability. Helps out with debugging and reasoning of the code :)

jelhan: I'm not sure if I get you right. Are we talking about not setting `responseBody` on `request` (Line 46-48) but constructing a new `request`. I think trying something like that was the root cause for this bug. Instead of passing the instance of `FakeRequest` through, a new pojo was constructed missing a lot of information. For a consumer it's hard to understand why this wrapper function about a route handler alters the route handler's interface. Constructing a new `FakeRequest` instance ourselves would be error prune and I'm not sure if it's even possible as pretender doesn't seem to export that class.

Alonski: We could do something like:
```js
let response = Object.assign({ requestBody: Object.assign({[...]: metadata}, data}, request)
```
Would this work?

jelhan: Not if we don't want to alter the interface. `request` is an instance of `FakeRequest` not a pojo.

gilest: I think that you can clone prototype objects using [Object.create](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create).

I think this snipppet works, and doesn't modify the original request.

```suggestion
const clonedRequest = Object.create(request);
clonedRequest.requestBody = Object.assign({ [file.key]: metadata }, data);

if (timedOut) {
  resolve(new Response(408));
  return;
}

resolve(fn.call(this, db, clonedRequest));
```

Screen Shot 2019-10-06 at 2 58 50 PM

tests/integration/components/mirage-handler-test.js
```diff
@@ -0,0 +1,42 @@
...
+  setupMirage(hooks);
+
+  test('upload handler passed schema and request.params throw', async function(assert) {
+    let self = this;
+    this.server.post('/folder/:id/file', uploadHandler(function(schema, request) {
+      assert.step('mirage-handler');
+
+      assert.equal(schema, self.server.schema, 'schema is provided');
+      assert.ok(typeof request.params === 'object', 'params property on request is an object');
+      assert.equal(request.params.id, '1', 'value of dynamic segment is present on params object');
```
gilest: Can you add an assertion for the requestBody properties?

```js
assert.deepEqual(request.requestBody, {
  'Content-Type': 'text/plain',
  file: {
    extension: 'txt',
    name: 'text.txt',
    size: 9,
    type: 'text/plain',
    url: 'data:text/plain;base64,c29tZS1kYXRh'
  }
});

```

addon/mirage/index.js
jelhan: Maybe we should not touch the request object at all but provide the file metadata as an additional argument to the route handler? 

```js
resolve(fn.call(this, ...arguments, metadata));
```

Alonski: Hmm. You have convinced me. Lets keep it your way for now. I generally don't like mutating objects if it can be avoided. Maybe here its for the best.

tests/integration/components/mirage-handler-test.js
jelhan: Done in a1aae2dfb41b3626c631cb0d640956b4ee22b394.

PR #176, opened 2018-10-03
Fixes #174 - Enable dropdown without upload

fran-worley:
Fixes #174 

@tim-evans I've tested this locally and it works, I'm not sure how you want me to test this as there aren't any component tests.

PR #275, opened 2019-09-28
helpful error message if invalid image is provided to mirage handler

jelhan:
Before this change test timed out if an invalid image was provided cause `img.onload` won't be called at all. Took me quite some time to understand what's happening. This PR adds a helpful error message.

It drops testing for `ember@2.16`. This is not done cause it's not working anymore with `v2.16` but because ember <= 2.18 had some issues with `Ember.onerror` and I was a little bit lazy trying to work-a-round that one. If you want to keep running the tests against 2.16, I can refactor to skip only the problematic test in 2.16.

This PR also consolidates the testing docs on GitHub pages and removes them from README. See #273

PR #276, opened 2019-09-29
Fixes #174 - Enable dropdown without upload (v2)

### LINKED ISSUE #174: Enable support for file-dropzone without file-upload
I have raised a [full issue](https://github.com/kaliber5/ember-bootstrap/issues/672) in the EmberBootstrap repository as I'm pretty sure that is the root cause of the issue, however I thought I would raise it for anyone else who uses both addons and incase you guys can shed some light on how the dropzone and file upload components work that could help me resolve the issue.

gilest:
This PR is a duplicate of #176 

I rebased this branch, but the build was still failing.

Seems to be related to this issue: ember-learn/ember-cli-addon-docs-yuidoc#4 

In this case it was fixed by adding a missing `@class` comment block.

Hope this is OK with you @fran-worley

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,6 @@
 /npm-debug.log*
 /testem.log
 /yarn-error.log
-/.vscode
 
 # ember-try
 /.node_modules.ember-try/
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@
 /npm-debug.log*
 /testem.log
 /yarn-error.log
+/.vscode
 
 # ember-try
 /.node_modules.ember-try/
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,8 +45,7 @@ jobs:
     # we recommend new addons test the current and previous LTS
     # as well as latest stable release (bonus points to beta/canary)
     - stage: "Additional Tests"
-      env: EMBER_TRY_SCENARIO=ember-lts-2.16
-    - env: EMBER_TRY_SCENARIO=ember-lts-2.18
+      env: EMBER_TRY_SCENARIO=ember-lts-2.18
     - env: EMBER_TRY_SCENARIO=ember-release
     - env: EMBER_TRY_SCENARIO=ember-beta
     - env: EMBER_TRY_SCENARIO=ember-canary
--- a/addon-test-support/index.js
+++ b/addon-test-support/index.js
@@ -5,9 +5,50 @@ export async function upload(selector, ...files) {
   let input = find(selector);
   assert(`Selector '${selector}' is not input element.`, input.tagName === 'INPUT');
   assert(
-    `All files must be instances of File/Blob type`,
+    'All files must be instances of File/Blob type',
     files.every((file) => file instanceof Blob)
   );
 
   return triggerEvent(input, 'change', { files });
+}
+
+export async function dragAndDrop(selector, ...files) {
+  let dropzone = find(selector);
+  assert(`Selector '${dropzone}' could not be found.`, dropzone);
+  assert(
+    'All files must be instances of File/Blob type',
+    files.every((file) => file instanceof Blob)
+  );
+
+  let dataTransfer = { files };
+
+  await triggerEvent(dropzone, 'dragenter', { dataTransfer });
+  await triggerEvent(dropzone, 'dragover', { dataTransfer });
+  return triggerEvent(dropzone, 'drop', { dataTransfer });
+}
+
+export async function dragEnter(selector, ...files) {
+  let dropzone = find(selector);
+  assert(`Selector '${dropzone}' could not be found.`, dropzone);
+  assert(
+    'All files must be instances of File/Blob type',
+    files.every((file) => file instanceof Blob)
+  );
+
+  let dataTransfer = { files };
+
+  return triggerEvent(dropzone, 'dragenter', { dataTransfer });
+}
+
+export async function dragLeave(selector, ...files) {
+  let dropzone = find(selector);
+  assert(`Selector '${dropzone}' could not be found.`, dropzone);
+  assert(
+    'All files must be instances of File/Blob type',
+    files.every((file) => file instanceof Blob)
+  );
+
+  let dataTransfer = { files };
+
+  return triggerEvent(dropzone, 'dragleave', { dataTransfer });
 }
--- a/addon-test-support/index.js
+++ b/addon-test-support/index.js
@@ -1,10 +1,13 @@
 import { find, triggerEvent } from '@ember/test-helpers';
 import { assert } from '@ember/debug';
 
-export async function upload(selector, file) {
+export async function upload(selector, ...files) {
   let input = find(selector);
   assert(`Selector '${selector}' is not input element.`, input.tagName === 'INPUT');
-  assert(`File must be instance of File/Blob type`, file instanceof Blob);
+  assert(
+    `All files must be instances of File/Blob type`,
+    files.every((file) => file instanceof Blob)
+  );
 
-  return await triggerEvent(input, 'change', { files: [file] });
+  return triggerEvent(input, 'change', { files });
 }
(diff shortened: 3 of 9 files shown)


gilest:
Good point. I'll try and write some tomorrow.

addon-test-support/index.js
```diff
@@ -1,10 +1,13 @@
...
   let input = find(selector);
   assert(`Selector '${selector}' is not input element.`, input.tagName === 'INPUT');
-  assert(`File must be instance of File/Blob type`, file instanceof Blob);
+  assert(
+    `All files must be instances of File/Blob type`,
+    files.every((file) => file instanceof Blob)
+  );
 
-  return await triggerEvent(input, 'change', { files: [file] });
+  return triggerEvent(input, 'change', { files });
```
jelhan: If I got it right, this adds support for testing upload of multiple files isn't it? Very helpful! :+1: Should document that. But as #275 rewrites the testing docs we might want to do it in another up following PR to avoid merge conflicts?

tests/integration/components/file-upload-test.js
```diff
@@ -0,0 +1,37 @@
...
+    this.setFile = (file) => {
+      assert.step(file.name);
+    };
+
+    await render(hbs`{{file-upload name="test" onfileadd=(action this.setFile)}}`);
+
+    await upload(
+      'input[type="file"]',
+      new File([], 'dingus.txt'),
+      new File([], 'dingus.png')
```
Alonski: Should these be in an array? Isn't that what the test helper uses?

jelhan: It uses the [rest parameter syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameter):

```js
export async function upload(selector, ...files) {
}
```

An awesome design choice IMO. It does not alter the existing interface and does not allow multiple types for the same argument. This should make documentation straight forward and even tooling support (if we provides types as some point of time).

gilest:
> > * In my test cases, adding a file without providing a `name` argument to the `file-upload` component throws a cryptic error. Maybe `name` should be documented as mandatory, with a helpful assertion for when implementers forget it.
> 
> I think the option has to be required as it's used for looking up the queue. Each file must be attached to a queue for upload and there isn't anything like a default queue. I also think that a default queue would make much sense in the architecture of this addon. It would be very error prune I guess. But as you suggested there should be a helpful assertion if `name` argument is missing. I think it's the same for `<FileDropzone>` isn't it?

I agree, probably for another PR.

Yes, `FileDropzone` will be the same, they both extend `BaseComponent` where this behaviour is defined.

> This is fixed as part of the testing docs rewrite done in #275.

Oh nice! I missed that from when the line-endings were messing up the diff.

It's a bit awkward having a local class called `File` when there is a browser-native class of the same name. Problem for another day though 😅

addon-test-support/index.js
gilest: Yeah, that's right. I figured multiple file selection is something we want to test since it's a documented feature.

It doesn't break the existing API though, which is nice.

If we can get #275 merged before this I'll update the docs in this PR.

addon/components/file-dropzone/component.js
```diff
@@ -68,14 +68,16 @@ const dragListener = new DragListener();
...
+  @yield {boolean} dropzone.valid
   @yield {Queue} queue
  */
 export default BaseComponent.extend({
 
   layout,
 
   supported,
+  active: false,
+  valid: true,
```
gilest: These properties were initially undefined. I think this was a violation of the public API (boolean). Also nice to have them explicitly defined here.

addon/components/file-dropzone/component.js
```diff
@@ -185,6 +187,9 @@ export default BaseComponent.extend({
         this[DATA_TRANSFER] = null;
       }
 
+      if (this.isDestroyed) {
+        return;
+      }
```
gilest: Tests occasionally failed with `Cannot call set on destroyed object`.

tests/integration/components/file-dropzone-test.js
```diff
@@ -0,0 +1,125 @@
...
+module('Integration | Component | file-dropzone', function(hooks) {
+  setupRenderingTest(hooks);
+
+  test('dropping a file calls ondrop', async function(assert) {
+    this.onDrop = (dataTransfer) => {
+      dataTransfer.files.forEach((file) => assert.step(file.name));
+    };
+    /**
+     Must set `allowUploadsFromWebsites=true` otherwise `FileDropzone.isAllowed` returns false
+    */
```
gilest: This kind of sucks, an alternative would be to detect if the component is being used in a test environment within `isAllowed`.

addon-test-support/index.js
```diff
@@ -1,10 +1,54 @@
...
+}
+
+export async function dragAndDrop(selector, ...files) {
```
gilest: I've kind of made a mess in here. Difficult to have a nice helper API and not duplicate a lot of code.

How much of this stuff do you think we should make available to users of the addon?

.travis.yml
```diff
@@ -45,7 +45,6 @@ jobs:
     # we recommend new addons test the current and previous LTS
     # as well as latest stable release (bonus points to beta/canary)
     - stage: "Additional Tests"
-      env: EMBER_TRY_SCENARIO=ember-lts-2.16
```
gilest: +1 for ceasing to support this version of ember.

Had trouble catching catching errors thrown by the `render` function using `setupOnerror`.

addon/components/file-dropzone/component.js
jelhan: I think we need to use `this.get()` in order to support Ember 2.18, isn't it?

addon-test-support/index.js
jelhan: What do you mean by mess? Looks clean and well structured to me. :+1:

.travis.yml
jelhan: Same issue faced in #275

addon/components/file-dropzone/component.js
gilest: You're right. I keep forgetting about `get`.

Fixed in d89dcb3 😅

addon-test-support/index.js
gilest: I guess I just mean the "boilerplate code" that is repeated.

.travis.yml
gilest: I guess the solution to this in both PRs is either of:

1. Skip tests that override `Ember.onerror` in `2.16` test runs
2. Drop support for `2.16`

@Alonski @jelhan which would you prefer?

.gitignore
```diff
@@ -16,6 +16,7 @@
 /npm-debug.log*
 /testem.log
 /yarn-error.log
+/.vscode
```
Alonski: Personally I don't think this should be ignored. `/.vscode` allows sharing project specific settings for VSCode. You can of course override them :)

tests/integration/components/file-dropzone-test.js
Alonski: This is possible no? Importing ENV?

gilest: Resolved in e7c85a7.

.gitignore
gilest: Removed in e7c85a7.

addon/components/file-dropzone/component.js
```diff
@@ -171,7 +153,10 @@ export default Component.extend({
   },
 
   isAllowed() {
-    return get(this[DATA_TRANSFER], 'source') === 'os' ||
+    const { environment } = getOwner(this).resolveRegistration('config:environment');
```
gilest: Not certain this is the correct way to check the `environment` of the consuming app.

Got this snippet from [ember-get-config readme](https://github.com/briarsweetbriar/ember-get-config).

Alonski: Asked on Discord :)
https://discordapp.com/channels/480462759797063690/491905849405472769/630290136982093844

Seems good!

PR #278, opened 2019-09-30
convert line endings CRLF -> LF

jelhan:
Noticed in #275 that some files are using `CRLF`. This is not inline with editorcofing configuration: https://github.com/adopted-ember-addons/ember-file-upload/blob/5056fc50754d455312c013a606042899b8433df7/.editorconfig#L8-L9

This changes line endings to `LF` for all files. You could verify that there isn't any other change with `git diff --ignore-space-at-eol`.


jelhan:
> Can you also add a lint rule for prettier to enforce this?

To be honest I have no idea how to set this up. Would be great if you could guide me to an example.

ESLint is already configured to check line endings but does that only on JavaScript files: https://github.com/adopted-ember-addons/ember-file-upload/blob/efba6e3330141da0803d13594817b4fc5c21ae9f/.eslintrc.js#L51

gilest:
Try a `.prettierrc` file containing

```json
{
  "endOfLine": "lf",
}
```

[Options reference](https://prettier.io/docs/en/options.html#end-of-line)

jelhan:
@gilest That would work but would not only check the file endings but a lot of other stuff also. Maybe we should just run prettier on the whole code base once and add it afterwards to CI? But that would be something for another PR, I guess.

jelhan:
Rebased to fix the merge conflict.

jelhan:
@Alonski Could you please consider merging this without a linting rule to prevent regression? We may at that one later but this is currently a blocker for #275, which is itself a blocker for #276.

jelhan:
Changed template lint configuration to also enforce unix linebreak style. Credits to @gilest for showing me that option on Discord. Thanks a lot!

Issue #273, opened 2019-09-28
Docs in readme and on GitHub pages are out of sync


jelhan:
The docs in README and GitHub pages are out of sync. At least for the cases I noticed the README is more up-to-date. I don't think it's a good idea to duplicate the docs. I would recommend to reduce the README to a simple usage example and update the docs on GitHug pages.

Alonski:
Awesome catch @jelhan . Can you please point to the exact issues? Also a PR would be very welcome

jelhan:
Fixed by #281

Issue #122, opened 2018-03-22
Use ember-cli-addon-docs for documentation


tim-evans:
The documentation is currently quite poor for this addon, with all of it in the README, and having a (partially) good website for this. This is something that will require a bit of work; the following tasks need to be done:

- [x] 🚀 Get ember-cli-addon-docs installed
- [ ] 📓 Fill out JSDocs for important areas of ember-file-upload
- [ ] ✅ Update tests to run against the documentation instead of the dummy app
- [ ] 🖊 Write documentation (like in the README)

raido:
@tim-evans Hi, I got something going over here - https://github.com/raido/ember-file-upload/tree/ember-cli-addon-docs

Basically just a default installation done and it boots up.

tim-evans:
Brilliant!

tim-evans:
Hi @samselikoff; we seem to be have some issues with the deploy here. I'm not particularly sure whom to get in touch with, but wanted to bring it to your attention.

Thanks for the utterly brilliant addon. This makes my life maintaining addons considerably easier and more pleasant. ❤️

raido:
It seems ember-cli-mirage 0.4.0+ is the culprit in prod builds. I downgraded mirage to 0.3.4 and it seems to work locally at least. PR -> https://github.com/tim-evans/ember-file-upload/pull/129

samselikoff:
@tim-evans et. al., missed this due to a backlog of gh notifications...

Let me know if running into more issues! Or pop into #ec-addon-docs in Slack

samselikoff:
Looks like you have an addon docs site up and running, sweet! 👊

gilest:
@Alonski I think this can be closed.

jelhan:
I think this one is resolved by the awesome work of @gilest in the last weeks.

PR #281, opened 2019-10-10
Update addon docs and remove docs from readme

gilest:
I have done a number of things here:

* Copy the newer recipes content (angle bracket examples) to the addon docs site (#273)
* Removed Configuration section (which was only in the readme file) outright – YUIDoc generated API docs should be single source of truth
* Use Ember FileUpload as the project name in readme and docs – it's a bit more of a modern project name without the double-curlies
* Removed YUIDoc comments from `BaseComponent` class since it's a private API
* Added all shared attributes to both `FileUpload` and `FileDropzone` component files, so that the API Reference section of the addon docs site can be the single source of truth for component attributes
* Improved the introduction a little bit. It doesn't make sense to talk about the addon as only providing a single component since #276 

One thing I haven't done yet is add a very simple usage example to the docs. In my opinion this isn't necessary right away, and we can backport it into the readme as we continue to improve the docs.

Also related: #122


gilest:
> I'm wondering if we still need the curly-invocation examples as there isn't anything special about using that components in curly-invocation.

I'm OK with using angle-bracket invocation in the examples. We could include a paragraph about how both invocation methods work.

Do you think I should update the component tests to use angle-bracket invocation also?

gilest:
> > Removed YUIDoc comments from `BaseComponent` class since it's a private API
> 
> Wouldn't the other components just inherit the docs (as the implementation) if we keep them on the `BaseComponent`?

Ideally but I'm having a few problems with this:

* I've tried using `@extends` and `@uses` but I can't get the arguments documented in `BaseComponent` to show on the child component API doc pages
* Do we want `BaseComponent` itself on the API docs pages since it shouldn't be used on its own? I tried using `@nodoc` but that had no effect. Can't find a YUIDoc API to hide it from docs page

Hopefully you're more familiar with YUIDoc and can help with these 😅 

## In `latest` branch

Screen Shot 2019-10-12 at 9 56 33 AM

`BaseComponent` has an API docs page

Screen Shot 2019-10-12 at 9 56 45 AM

`FileUpload` is missing its inherited arguments

## With these changes

Screen Shot 2019-10-12 at 10 03 20 AM

* `BaseComponent` no longer has an API docs page
* `FileUpload` and `FileDropzone` pages have all documented arguments

README.md
```diff
@@ -1,350 +1,21 @@
...
-- [Installation](#installation)
-- [Configuration](#configuration)
-- [Recipes](#recipes)
-- [Integration](#integration)
-- [Access to the global list of uploading files](#access-to-the-global-list-of-uploading-files)
-- [Acceptance Tests](#acceptance-tests)
-- [S3 Direct uploads](#s3-direct-uploads)
-- [Contributing](#contributing)
-- [Publishing](#publishing)
+To use the uploader, you must provide a name (for proper queueing and bundling of resources), and an upload URL.
```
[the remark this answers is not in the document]
gilest: I agree with you, removed in 37d40d5.

jelhan:
> Hopefully you're more familiar with YUIDoc and can help with these

To be honest: I'm not. @Alonski could you help here?

Alonski:
I actually also have had issues with YUIDocs in the past :(
Do you think we should merge this anyways? Possibly fix further issues in later PRs?

addon/components/file-dropzone/component.js
```diff
@@ -80,6 +80,51 @@ export default BaseComponent.extend({
   active: false,
   valid: true,
 
+  /**
+    Whether multiple files can be selected when uploading.
+    @argument multiple
+    @type {boolean}
+   */
+  multiple: null,
```
jelhan: If we have to duplicate the documentation to make YUIDocs happy, do we also have to duplicate the default values? If so we should remove them (`multiple`, `name`, `disabled`, `accept`, `onfileadd`) from `BaseComponent` as they are overriden by both child classes.

package.json
```diff
@@ -39,6 +39,7 @@
   "devDependencies": {
     "@ember/optional-features": "0.7.0",
     "broccoli-asset-rev": "2.7.0",
+    "ember-angle-bracket-invocation-polyfill": "^2.0.2",
```
jelhan: We only need this one for tests, right? Otherwise this has to be moved from `devDependencies` to `dependencies`.

gilest: I think that Ember apps lower than `3.10` would install `ember-angle-bracket-invocation-polyfill` into their own app dependencies.

That way there is no confusion about where angle bracket support comes from in their apps. 

This makes the most sense to me. What do you think?

jelhan: But would it be required if app uses curly component invocation syntax? I don't think so because angle bracket component invocation is not used in any of the component templates. But if I'm wrong we should document that requirement.

addon/components/file-dropzone/component.js
gilest: Staged in #287

package.json
gilest: I don't think `ember-angle-bracket-invocation-polyfill` is a requirement for consuming apps unless:

- The consuming app uses Ember < `3.10`
- AND invokes *any* component using angle-bracket syntax

PR #292, opened 2019-11-15
Update component docs to use angle bracket syntax

mcfiredrill:
I tried use ember-angle-brackets-codemod, but it didn't seem to work on
.js files. So I edited the doc comments manually.


addon/components/file-dropzone/component.js
```diff
@@ -39,15 +39,15 @@ const dragListener = new DragListener();
         {{#if dropzone.supported}}
           Drag and drop images onto this area to upload them or
         {{/if}}
-        {{#file-upload name="photos"
-                      accept="image/*"
-                      multiple=true
-                      onfileadd=(action "uploadImage")}}
+        <FileUpload name="photos"
+                    accept="image/*"
+                    multiple=true
```
jelhan: `multiple=true` is not valid HTML, is it? I think it should be `<FileUpload multiple />`.

addon/components/file-dropzone/component.js
```diff
@@ -39,15 +39,15 @@ const dragListener = new DragListener();
         {{#if dropzone.supported}}
           Drag and drop images onto this area to upload them or
         {{/if}}
-        {{#file-upload name="photos"
-                      accept="image/*"
-                      multiple=true
-                      onfileadd=(action "uploadImage")}}
+        <FileUpload name="photos"
+                    accept="image/*"
```
jelhan: I fear we don't support passing HTML attributes to `<FileUpload>` yet. It doesn't seem to pass them to `<input>`. At least I don't see a `...attributes` in the template.

addon/components/file-dropzone/component.js
```diff
@@ -39,15 +39,15 @@ const dragListener = new DragListener();
...
+                    multiple=true
+                    @onfileadd={{action "uploadImage"}}>
```
jelhan: The coding style looks different than what I would expect. I'm used to this one:

```hbs
<FileUpload
  @onfileadd={{action "uploadImage"}}
  name="photos"
  accept="image/*"
/>
```

So it's basically about this rules:

- If there are multiple lines, each attribute or property should be on a new line
- It should be intended by two spaces.
- Arguments should be before HTML attributes.

I thought that's also what's used in Ember docs but noticed that Ember docs aren't consistent on these ones.

knownasilya: I usually do html attributes first and then args, usually actions last

mcfiredrill:
> I fear we missed some work to fully support angle bracket invocation. I see two ways to address it:

> Do not pass HTML attributes but arguments on component invocation or
> fix missing attributes forwarding before merging this one.

Sounds good, I'm using this addon with angle brackets in my app now so I will make another PR with those fixes before getting this merged.

addon/components/file-dropzone/component.js
mcfiredrill: Hmm I wonder if there is any [ember-template-lint rule](https://github.com/ember-template-lint/ember-template-lint) about this...

addon/components/file-dropzone/component.js
```diff
@@ -39,15 +39,15 @@ const dragListener = new DragListener();
         {{#if dropzone.supported}}
           Drag and drop images onto this area to upload them or
         {{/if}}
-        {{#file-upload name="photos"
-                      accept="image/*"
-                      multiple=true
-                      onfileadd=(action "uploadImage")}}
+        <FileUpload name="photos"
```
mcfiredrill:
```suggestion
        <FileUpload @name="photos"
```

addon/components/file-dropzone/component.js
knownasilya: ...attributes are applied to the parent element unless tagName is set to a empty string.

gilest:
See [file-upload-test.js](https://github.com/adopted-ember-addons/ember-file-upload/blob/master/tests/integration/components/file-upload-test.js) for the angle-bracket API that is currently covered by integration tests.

It's not an absolutely ideal API but it does allow us to support a wide range of ember versions.

addon/components/file-dropzone/component.js
Alonski: @mcfiredrill I am glad you asked 😄 
https://github.com/ember-template-lint/ember-template-lint/pull/871
I requested this a while ago :)
https://github.com/ember-template-lint/ember-template-lint/issues/684

Although it isn't exactly what @knownasilya mentioned

Alonski:
@Turbo87 gave me a few tips on this:
```
I would prefix everything with @
if you don't then the fact that the ...attributes element is the <input type="file"> becomes public API and I'm not sure you'll want that in this case
I generally prefer a defined public API instead of the "pass everything to some child element"
...attributes should still be on the top-level element though so that CSS classes can be applied from the outside, to e.g. add margins on the component
```
```
Hmm so we should explicitly use @ and then inside the component use <input name={{@name}}> or something?
```
```
yeah, I think so
```

mcfiredrill:
@Alonski Very good points! I didn't think about the <input> basically becoming public API.

It seems like I got really hung up on getting the component to work with normal HTML attributes without really thinking about the tradeoffs or alternatives.

https://github.com/adopted-ember-addons/ember-file-upload/pull/292#pullrequestreview-317520352
> I fear we missed some work to fully support angle bracket invocation. I see two ways to address it:

>1 . Do not pass HTML attributes but arguments on component invocation or
> 2. fix missing attributes forwarding before merging this one.

Maybe @jelhan 's initial suggestion 1 here was a better idea after all. 🤔

mcfiredrill:
@gilest Thank you, I didn't even realize that was in the tests! 😅

jelhan:
@Alonski I'm not sure if I agree. Designing this component from scratch using an outer-html approach, would we really end up with this as best API?

```hbs
<label ...attributes>
  <input>
</label>
```

I've the feeling that we would have done something like:

```hbs
{{#if (has-block)}}
  {{yield
    (hash
      fileChanged=(action "change")
    )
  }}
{{else}}
  <label
    for={{inputId}}
  >
  <input
    id={{inputId}}
    type="file"
    onchange={{action "change"  value="target.files"}}
   ...attributes
  >
{{(/if}}
```

Turbo87:
@jelhan if you implement it that way you would have no way of controlling the margins around the component from the outside without adding another unnecessary wrapper component. I would consider the fact that the component is using a native input element inside an implementation detail that should not be relied upon.

mcfiredrill:
> Hmm so we should explicitly use @ and then inside the component use `<input name={{@name}}>` or something?

using `@` for all arguments seems to be supported already according to this test:
https://github.com/adopted-ember-addons/ember-file-upload/blob/master/tests/integration/components/file-upload-test.js

So I think if I just change all the arguments in the docs to be `@` prefixed, would this PR be merge-able?

>  `<input name={{@name}}>` 

I think we could leave the templates as is without `@` prefixing the arguments in the template. This will keep them backwards compability with `{{}}` curly components, if I understand correctly?

I renamed my related PR to "Support html attributes in angle bracket components", since angle bracket components technically already work.

mcfiredrill:
> So I think if I just change all the arguments in the docs to be @ prefixed, would this PR be merge-able?

Done in https://github.com/adopted-ember-addons/ember-file-upload/pull/292/commits/aa98e71541a82b5344b63c71b65e86efe44f9a45

Hmm build is failing...I haven't changed any code in this PR however. Someone want to try running it again for me? 🛠

mcfiredrill:
bump...anyone know why my build failed? :)

Alonski:
@mcfiredrill Restarted. Ping me on Discord if it fails again please 🙂

gilest:
@mcfiredrill Hello from the future. Care to rebase this so the CI can run again?

mcfiredrill:
Hi there, it's been awhile! I just merged master into this branch.

gilest (approved):
LGTM.

@jelhan Your requested changes had been addressed – yes?

addon/components/file-upload/component.js
```diff
@@ -26,16 +26,16 @@ import uuid from '../../system/uuid';
       <label for='name'>
       {{input type='string' value=changeset.name id='name'}}
 
-      {{#file-upload name="avatar"
-                     accept="image/*"
-                     onfileadd=(action 'setAvatar' changeset)}}
+      <FileUpload @name="avatar"
+                  @accept="image/*"
+                  @onfileadd={{action 'setAvatar' changeset}}
```
gilest, line 31:
```suggestion
                  @onfileadd={{action 'setAvatar' changeset}}>
```

addon/components/file-dropzone/component.js
```diff
@@ -41,15 +41,15 @@ const dragListener = new DragListener();
...
         {{/if}}
-        {{#file-upload name="photos"
-                      accept="image/*"
-                      multiple=true
-                      onfileadd=(action "uploadImage")}}
+        <FileUpload @name="photos"
+                    @accept="image/*"
+                    @multiple=true
+                    @onfileadd={{action "uploadImage"}}>
           <a id="upload-image" tabindex=0>Add an Image.</a>
```
gilest, line 48:
```suggestion
        <FileUpload @name="photos"
                    @accept="image/*"
                    @multiple=true
                    @onFileAdd={{this.uploadImage}}>
          <a id="upload-image" tabindex=0>Add an Image.</a>
```

addon/components/file-upload/component.js
gilest, line 31:
```suggestion
      <FileUpload @name="avatar"
                  @accept="image/*"
                  @onFileAdd={{this.setAvatar}}>
```

gilest (changes requested):
Since (#476) `onfileadd` has been deprecated in favour of `onFileAdd`.

Also found a `>` was missing.

Also suggest we:
- Drop `action` helper from examples
- Use `fn` helper to pass arguments to callbacks

Please rebase 😅

mcfiredrill:
Sorry for the delay, I have merged master.

> Also suggest we:

> Drop action helper from examples
> Use fn helper to pass arguments to callbacks

I could try to address these here, or that could be another PR?

gilest:
> > Drop action helper from examples
> > Use fn helper to pass arguments to callbacks
> 
> I could try to address these here, or that could be another PR?

Looks like you addressed these in 71d3298b2d6daa00d416cd51f52fb668f1b1f749

gilest (approved):
LGTM. Believe this syntax is correct for the current version.

gilest:
@jelhan believe this can be merged now. Let me know if you still need changes

PR #293, opened 2019-11-19
Support html attributes in angle bracket components

mcfiredrill:
Support for angle bracket components.
docs: https://github.com/adopted-ember-addons/ember-file-upload/pull/292

I know at the least `...attributes` needs to be added to the template, I'm still working out if any other code can be removed or if anything else is needed.


addon/components/file-upload/template.hbs
```diff
@@ -7,6 +7,7 @@
   disabled={{disabled}}
   onchange={{action "change" value="target.files"}}
   hidden
```
mcfiredrill: Can these other attributes be removed in this template since they are now passed in as regular HTML attributes?
```
  accept={{accept}}
  capture={{capture}}	 
  multiple={{multiple}}	
  disabled={{disabled}}	 
```
Or perhaps they can just be given defaults in the template and then the `...attributes` will override them?

jelhan: This would be a breaking change as we won't support curly bracket invocation syntax anymore but I think that's fine as we prepare a new major anyways. @Alonski @gilest any thoughts on that one?

mcfiredrill:
>As @knownasilya pointed out here #292 (comment) the components needs to be refactored to tagless components cause otherwise HTML attributes are applied to the tag defined by them as well.

Sounds good. Seems like I could move the `label` for file-upload directly into the template. 

Will need to refactor away from using `attributeBindings` as well.

mcfiredrill:
> Sounds good. Seems like I could move the label for file-upload directly into the template.

Is it necessary for the label to wrap around the entire template, as in the current implementation? 
If not, I have a feeling this check for VALID_TAGS could be removed.
https://github.com/adopted-ember-addons/ember-file-upload/blob/master/addon/components/file-upload/component.js#L160

jelhan:
> Is it necessary for the label to wrap around the entire template, as in the current implementation?
> If not, I have a feeling this check for VALID_TAGS could be removed.
> https://github.com/adopted-ember-addons/ember-file-upload/blob/master/addon/components/file-upload/component.js#L160

Oh, I wasn't aware of this. The main question seems to be: Should we support changing the tag of `<label>` at all? If so, that might be a show-stopper. If I didn't missed something that's not yet supported in template only components, is it?

I think we should ship valid HTML and drop that use case at all. Valid HTML would also be not wrapping input by label but assigning them to each user using `for` attribute on label.

@Alonski, @gilest Any thoughts?

mcfiredrill:
FWIW this is already looking to be a breaking change.

> I think we should ship valid HTML and drop that use case at all. Valid HTML would also be not wrapping input by label but assigning them to each user using for attribute on label.

👍 for valid HTML.

mcfiredrill:
That said I think I don't 100% understand why the component was implemented as a `label` in the first place.

addon/components/file-upload/template.hbs
gilest: The current implementation is classic component classes which support the angle bracket invocation syntax.

Important to remember that these are separate concepts in ember:
* **component type/class** (classic or glimmer)
* **component invocation** (curly or angle)

I think that classic components are a more flexible implementation for now since it supports both new and older ember apps.

If we use "tagless" glimmer components I think we may have difficulty supporting older versions of ember.

Is there any functionality or API that we are missing out on from not using `...attributes`?

gilest:
YeahI think the `<label>` element that wraps the `<input>` is an interactive (clickable, and maybe draggable) area in many browsers.

We should make sure we follow the spec closely here for usability and accessibility.

gilest:
@mcfiredrill Love the enthusiasm you're bringing 👍 

But this is a pretty long-running addon, and I think supporting legacy apps and classic components is something we should keep doing for a while yet.

mcfiredrill:
@gilest Thanks a lot for your feedback!

I haven't looked at it much yet so I'm going to investigate what benefits using ember-angle-bracket-invocation-polyfill in dependencies would bring.

I am thinking to maybe follow this advice, and not worry about supporting html attributes for now.
https://github.com/adopted-ember-addons/ember-file-upload/pull/292#issuecomment-557821347

I may open a new PR or just totally redo this one.

addon/components/file-upload/template.hbs
jelhan: @gilest I fear you are missing inner-html vs outer-html in your concepts. Components using inner-html (`@ember/component` without `tagName: ''`) do not play well together with `...attributes` cause the HTML attributes passed on invocation would be set on the the element defined by component _and_ on the elements that have `...attributes`.

E.g. if you have a component `<FormElement>` like this:

```js
export default Component.extend({
  tagName: 'label',
});
```

```hbs
<input ...attributes>
```

And you invoke it with `<FormElement required />` the rendered HTML would be:

```html
<label required>
  <input required>
</label>
```

That's not what you want in most cases.

gilest: You're right that's an important distinction! I think `tagName: ''` was kind of discouraged but I've used it a lot myself.

gilest:
> @gilest Thanks a lot for your feedback!
> 
> I haven't looked at it much yet so I'm going to investigate what benefits using ember-angle-bracket-invocation-polyfill in dependencies would bring.
> 
The polyfill is currently included in `devDependencies` so that we can test angle bracket invocation in out integration specs.

Moving it to `dependencies` would mean that users of the addon will get it too. This would allow us to use (while still supporting older versions of ember) angle bracket syntax within the addon code itself (rather than just the tests).
>
> I am thinking to maybe follow this advice, and not worry about supporting html attributes for now.
> [#292 (comment)](https://github.com/adopted-ember-addons/ember-file-upload/pull/292#issuecomment-557821347)
> 
> I may open a new PR or just totally redo this one.
Both options sound like a good approach 👍

mcfiredrill:
Hmm I'm confused, I see the link on github (actually a 404) saying the build failed.
But I found the build on travis just by clicking around there and it seemed to have passed:
https://travis-ci.org/github/adopted-ember-addons/ember-file-upload/builds/774490653

gilest:
@mcfiredrill Please push again – we're on GitHub actions now 😅

mcfiredrill:
@gilest Nice! Merged master.

mcfiredrill:
Tests are passing but I'm trying to remember exactly what I was trying to do with this PR (nearly 3 years ago!), and it looks like I wanted to re-do at least a large portion of it.

> I may open a new PR or just totally redo this one

gilest:
Both FileDropzone and FileUpload implement splattributes since #514 which has been released as 4.1.0-beta.0

PR #294, opened 2019-12-23
fix: Remove DisableGPU from Testem

Alonski:


jelhan:
I've restarted the build for latest `master` just to double check if the error is related to this change or not: https://travis-ci.org/adopted-ember-addons/ember-file-upload/builds/612303498?utm_source=github_status&utm_medium=notification

jelhan:
I think the issue with tests has been fixed on master. Still think this change is value as it updates `testem.js` to be inline again with blueprint: https://github.com/ember-cli/ember-addon-output/blob/master/testem.js But could also do it as part of a general upgrade of Ember CLI blueprints. @Alonski it's up to you in my opinion either to rebase or to close this one.

gilest:
Restarting builds. If they pass I suggest we merge and move on

jelhan:
> Restarting builds. If they pass I suggest we merge and move on

I think you need to rebase.

PR #305, opened 2020-02-27
Update dependencies in attempt to fix Ember Try test builds

gilest:
Travis CI builds have been failing since 5bb1d65 (29 October 2019).

These builds are failing because Ember Try deletes the `yarn.lock` file and installs fresh dependencies.

It seems like a dependency within our dependency tree has had a minor update that has broken a few acceptance tests.

I have struggled to find the specific dep that is responsible for this. But I *think* updating this combination of deps does work. Hopefully the build passes 😅 

Updated dependency `ember-cli-addon-docs` to `0.6.16`.

Updated dependency `ember-cli-mirage` to `1.1.6`.

Use Pretender's `fake-xml-http-request` package, and include `ember-auto-import` in devDependencies so that it can be imported in a unit test.

Questions:
* Is this the right methodology for importing a utility class into an addon's tests?
* I know updating multiple dependencies at once isn't ideal – any ideas about how we could validate that this build does not break anything? What is not covered by tests that ought to be?


gilest:
I don't think any public APIs should have changed.

I was a bit wary of the `ember-cli-mirage` update, but there is quite good test coverage for the mirage handler that we provide and all of those test pass.

`ember-cli-addon-docs` should be ok – we can monitor the docsite and fix any problems that might crop up when we release.

Alonski:
Ok sounds good! Do you have permissions to merge and publish?

jelhan:
Thanks a lot for working on this one. Looks good to me.

> - Is this the right methodology for importing a utility class into an addon's tests?

Haven't reviewed why the dependency is needed in the first case but importing it explicitly is definitely better than relying on a global.

> - I know updating multiple dependencies at once isn't ideal – any ideas about how we could validate that this build does not break anything? What is not covered by tests that ought to be?

I don't see a big risk here especially cause the next release will be a major one as `master` already contains breaking changes.

jelhan:
Hope it was fine that I pushed the merge button. If I'm right everyone of us three should have permissions to merge for this repo.

I will try to provide release-it + lerna-changelog setup by the end of today. Any breaking change left that we should include in next master?

gilest:
> Haven't reviewed why the dependency is needed in the first case but importing it explicitly is definitely better than relying on a global.

The `FakeXMLHttpRequest` used to be a global provided by `ember-cli-mirage`/`pretender`.

It is now provided as a separate package that is imported by `pretender`.

`$ yarn list`
```
├─ pretender@2.1.1
│  ├─ fake-xml-http-request@^2.0.0
```

It has always been in this project's dependency tree, I have just added it explicitly to devDependencies since we now import it directly.

Issue #238, opened 2019-04-28
Remove Observer Usages


Alonski:
Currently, there are two usages of `observers` in this addon:
1. https://github.com/adopted-ember-addons/ember-file-upload/blob/latest/addon/services/file-queue.js#L78
2. https://github.com/adopted-ember-addons/ember-file-upload/blob/e63a1a9ab05fb34001c1207316f94511556e3fca/addon/queue.js#L158

These files seem oddly similar as well. Maybe we don't need both?

Need to remove these observer usages as observers are less recommended to be used.

gilest:
Removed in #307

PR #307, opened 2020-02-27
Remove observers

gilest:
Had a go at this re: #238.

Up to you folks whether this is better than keeping an Observer.

In this implementation I first extracted some duplicate code into a Mixin. I know Mixins are EOL too, but I think they are still better than duplicate code for now.

I then replaced the `flushFilesWhenSettled` observer with a plain function called `flush`.

Note: The FileQueue service tracks files accross all queues. The Queue class tracks its own files.

We only need to call `flush()` on the Queue class. It will check if it is has a `FileQueue` service and call `flush()` on it.

The only places where files have their `state` property changed is in the `upload()` function within the File class, and in the FileQueue service test.

### DIFF
--- a/addon/file.js
+++ b/addon/file.js
@@ -102,6 +102,7 @@ function upload(file, url, opts, uploadFn) {
     set(file, 'state', 'failed');
     return RSVP.reject(error);
   }).finally(function () {
+    get(file, 'queue').flush();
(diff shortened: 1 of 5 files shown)


gilest:
I agree. It's kind of painful because I think the architecture is a little worse after this change 🤷‍♀

PR #320, opened 2020-03-09
do not pin dependencies

jelhan:
This does two changes:

1. Tell RenovateBot to not pin any dependencies but use the existing lock file.
2. Unpin all dependencies and run `yarn upgrade`.

Pinning the dependencies does not make much sense in my opinion if we have a lockfile:

- The yarn lockfile does not only pin direct but also indirect dependencies. There isn't much value from pinning them also in `package.json`.
- Pinning a dependency breaks `yarn upgrade`. To upgrade a dependency way more work is required and it's more noisy as even forward-compatible upgrades require a change to `package.json`.

Closes #310 and #317


jelhan:
@Alonski Do you mind reviewing as you have set up the Renovate initially?

gilest:
I'm not sure I have an opinion on this. Interested to hear more pros and cons though.

jelhan:
Closing and reopen in hope to fix Travis by it. It seems to have trouble checking out the branch.

jelhan:
Not sure what happens to `continuous-integration/travis-ci/push` task but as `continuous-integration/travis-ci/pr` is passing for the last commit, merging this one should be safe.

Alonski:
Just had time to look now. Good job!

Issue #306, opened 2020-02-27
Roadmap to v3


jelhan:
Lets make sure that we don't miss a breaking change that should be included in the next major release. `master` already includes breaking changes and wasn't released for days. So we should have that major rather sooner than later.

- [x] Drop support for node 8, which is EOL. #308 
- [x] Automate releases with release-it and lerna-changelog. #309 
- [ ] ~Decide on target for splat attributes (`...attributes`)~ #293
- [x] Drop support for IE 11 #319 
- [x] Rename `upload` test helper to `selectFiles` #322 
- [x] Drop support for legacy import at `<app-name>/tests/helpers/upload` #322

gilest:
I'm getting to work on node 8 support now.

jelhan:
Should consider dropping support for IE11 to reduce the maintenance work. Market share isn't high anymore and seems like no one wants to invest into supporting it. At least some open issues which haven't been addressed for some time.

gilest:
### IE11

The problem for me is that it's difficult to develop and test in IE11 in my environment. If someone is willing to own it that'd be great.

### Splat attributes
From my perspective a lot of the internals of this addon are using "legacy" architecture.

As long as we can continue to provide backwards compatibility I think slowly modernising the internal code is a good idea.

In this case maybe we could rearchitect the internal component invocation by moving `ember-angle-bracket-invocation-polyfill` from devDependencies to dependencies.

There is a cost to doing this though, which is that we would be installing the polyfill in **all consuming applications**.

Looking at the polyfill docs maybe that isn't so bad:
* Supports Ember 2.12 -> 3.9
* Completely inert when running Ember 3.10 or higher

jelhan:
> As long as we can continue to provide backwards compatibility I think slowly modernising the internal code is a good idea.
> 
> In this case maybe we could rearchitect the internal component invocation by moving `ember-angle-bracket-invocation-polyfill` from devDependencies to dependencies.

We could also consider dropping support for Ember < 3.12.

> From my perspective a lot of the internals of this addon are using "legacy" architecture.

I agree. Maybe that's simply to much for doing just now. Maybe we should release a v3 without touching that part to not block the release even longer. And then work on v4 which addresses the splat argument and other points.

jelhan:
Another breaking change that we might want to include in v3 is renaming event handlers to `camelCase`. See #284. Not as complex as splat attributes but would align the addon more with the rest of the ecosystem.

[...]

PR #322, opened 2020-03-10
Renamed `upload` test helper to `selectFiles` and dropped support for legacy import at `app/tests/helpers/upload`

gilest:
Is `test-support/helpers/upload.js` safe to delete?

#306


jelhan:
> Is `test-support/helpers/upload.js` safe to delete?

Yes I think so. But should be mentioned in the title so that it will be included in changelog directly.

gilest:
> Yes I think so. But should be mentioned in the title so that it will be included in changelog directly.

See that you've done that for me, thanks 👍

PR #327, opened 2020-03-10
Improve YUIDoc comments and fix build warnings

gilest:
Was seeing a few YUIDoc warnings in my console.

Added missing `@method` tags in `WithFilesMixin` and `Queue`.

Also removed a comment documenting the `init()` method in `FileQueueService`. My reasoning for this is.

- `@constructor` tag is meant to be used in a class block only - it's invalid YUIDoc
- If I tag this code block with `@method` then the `init()` method will show up on the docs site
- In my opinion this comment is superfluous

### DIFF
--- a/addon/mixins/with-files.js
+++ b/addon/mixins/with-files.js
@@ -25,7 +25,7 @@ export default Mixin.create({
     Files *may* be requeued by the user in the `failed` or `timed_out`
     states.
 
-    @type {function}
+    @method flush
    */
   flush() {
     let fileQueue = get(this, 'fileQueue');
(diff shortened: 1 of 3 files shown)


jelhan:
Ups, this should have been part of `3.0.0` isn't it? Guess we just have a patch release soon.

gilest:
It's quite a subtle change so not important to cut a new release IMO

jelhan:
> It's quite a subtle change so not important to cut a new release IMO

But on the other hand it's not much work thanks to the automation. :smile:

And at least for me doing small releases is even less work as you don't have to worry to miss something important. If you are sure what will be bundled in it's just a serious of `release-it`, hitting enter a few times and entering OTP for NPM.

TL;DR: Released v3.0.1. :tada:

Issue #284, opened 2019-10-28
Rename events to camelCase


jelhan:
Events are currently written in lowercase: `onfileadd`, `ondrop`, `ondragenter`, `ondragleave`.

As far as I'm aware camelCase event names gets more and more a shared convention in ember ecosystem. Even Ember Power Select switched it's API in the latest major release to camelCase event names.

I think we should do the same and rename to: `onFileAdd`, `onDrop`, `onDragEnter`, `onDragLeave`.

If we don't want to introduce a breaking change, it could be done with [`deprecatingAlias`](https://api.emberjs.com/ember/3.13/functions/@ember%2Fobject%2Fcomputed/deprecatingAlias).

gilest:
Closed by #328

PR #328, opened 2020-03-11
Renamed all events to be camelCase

gilest:
This is a breaking change to the public API to bring event names more in line with convention.

Requested in #284 

As @jelhan mentioned, we could use [deprecatingAlias](https://api.emberjs.com/ember/3.13/functions/@ember%2Fobject%2Fcomputed/deprecatingAlias) to support both names for a time.

### DIFF
--- a/addon/components/base-component.js
+++ b/addon/components/base-component.js
@@ -14,7 +14,7 @@ export default Component.extend({
   didReceiveAttrs() {
     this._super(...arguments);
     if (get(this, 'queue')) {
-      setProperties(get(this, 'queue'), getProperties(this, 'accept', 'multiple', 'disabled', 'onfileadd'));
+      setProperties(get(this, 'queue'), getProperties(this, 'accept', 'multiple', 'disabled', 'onFileAdd'));
     }
   },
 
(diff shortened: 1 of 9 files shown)


gilest:
Since we're waiting for publish permissions, we could possibly include this API change in v3 also. #306

gilest:
Yeah we would you prefer to:

- Maintain a separate branch with `breaking` changes 
- Use `deprecatingAlias` and similar features to warn about API changes coming in v4

jelhan:
> Which would you prefer?
> 
>     * Maintain a separate branch with `breaking` changes
>     * Use `deprecatingAlias` and similar features to warn about API changes coming in v4

I think the discussed changes are to fundamental to support both APIs in parallel. At least that would be to much work that could be better spend at other parts.

Deprecating parts of the old API will just cause churn cause consumers might now address the deprecation but will most likely change it again on `v4` when other parts are refactored.

So I tend to create a `dev-v4` branch now and target all breaking changes against that one. We should just make sure that we don't stay to long with both branches as porting pull requests to `master` and `dev-v4` is also causing some overhead.

To reduce the additional work by maintaining `master` and `dev-v4` we could start right now to only do hot fixes on `v3` and bundle even non-breaking enhancements for `dev-v4`.

gilest:
Closed in favour of #476

PR #330, opened 2020-03-12
add an upgrade guide for v3 to changelog

jelhan:
Breaking changes are not that clearly structured. I think a few upgrade notes will help. Not sure if I covered everything and to be honest haven't actually done an upgrade myself yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ The new `selectFiles` test helper expects a selector as first and one or many fi
 await selectFiles('#selector', new File([], 'empty.txt'));
 ```
 
+Please find more details in the [updated documentation for testing](https://adopted-ember-addons.github.io/ember-file-upload/docs/testing).
+
 #### :boom: Breaking Change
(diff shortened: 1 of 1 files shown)


gilest:
Agree about structure. Will be more careful with PR names and labels now that I understand how they end up in the changelog.

CHANGELOG.md
```diff
@@ -1,17 +1,40 @@
...
+import upload from '<app-name>/tests/helpers/upload';
+```
+
+Please change both of them to:
+
+```js
+import { selectFiles } from 'ember-file-upload/test-support';
+```
+
+The new `selectFiles` test helper expects a selector as first and one or many files as additional arguments. It does not take a file name as an argument. Please set the file name on the `File` instance:
```
gilest: Could possibly link to the docs here.

jelhan: :+1:

jelhan:
> Will be more careful with PR names and labels now that I understand how they end up in the changelog.

It's always difficult. Especially cause it not easy to split pull requests later to have them fit better in a changelog. I think adding additional information to the changelog afterwards is totally fine.

Issue #332, opened 2020-03-13
Documentation page is broken on Microsoft Edge


ndekeister-us:
Hello,

When we try to access documentation page with Microsoft Edge (Version: 18.18362) we have an error

![ember-file-upload-edge](https://user-images.githubusercontent.com/56396753/76638325-8d412c80-654c-11ea-890d-363c0da2c7aa.jpg)

(We have same error with v17 or v16)

Reproduced locally using `ember s` on `master` branch

jelhan:
Thanks for reporting. Seems like Edge is not listed explicitly in the list of target browsers: https://github.com/adopted-ember-addons/ember-file-upload/blob/8fb37d8cbefbd13c8103064a668e15bfc173a0b5/tests/dummy/config/targets.js#L3-L7

Do you have time to test if adding it resolves the issue?

Documentation for JS transpiling and build target configuration could be found [here](https://guides.emberjs.com/release/configuring-ember/build-targets/) and [here](https://cli.emberjs.com/release/advanced-use/asset-compilation/).

ndekeister-us:
Yes, adding `last 1 edge versions` solve it, thanks

Created #333

jelhan:
Fixed by #333 and released as `v3.0.2`.

ndekeister-us:
Just wondering if the documentation page is auto-updating with the last release? (if yes, it is style not accessible via Edge, with the same error)

jelhan:
It's deployed. Latest version is updated accordingly. But can confirm that it's still broken for Edge. Also seeing a syntax error in console. Guess we need to investigate what's going on. Reopening.

jelhan:
Facing a similar issue in one of my apps. Syntax error is thrown in Edge after removing IE11 from build targets. In both cases not transpiled object spread syntax seems to cause the error.

jelhan:
Released v3.0.3, which fixes this issue finally. Docs should be usable in Microsoft Edge again. Please reopen if you still face issues.

ndekeister-us:
Thanks, tested on Edge15-18 and it is ok 🙏

PR #333, opened 2020-03-13
Add edge to build targets for docs app

ndekeister-us:
See #332 

Add `last 1 edge versions` to dummy app targets so that generated code is compatible with Microsoft Edge.

Before
ember-file-upload-edge

After
Capture d’écran 2020-03-13 à 17 44 54


jelhan:
Thanks a lot. Will try to do a patch release later today.

jelhan:
Published `v3.0.2` including this fix.

PR #355, opened 2020-04-29
fix: allow uploading a File without a Queue

### LINKED ISSUE #351: File#queue is undefined
Upon upgrading to the `v3.0.3`, our test quite is reporting an error from this line

https://github.com/adopted-ember-addons/ember-file-upload/blob/4469a3e36f64167940bac65978b57d1bbd936d1c/addon/file.js#L105

In our instance, `file.queue` is `undefined`, and thus `flush` cannot be called on it.

I have been finding a hard time determining where `File#queue` is defined at all -- it does not seem to be part of the definition of the class.

Is there something we might be missing that's causing this issue? With the previous version -- `2.7.1` -- the tests were working fine. Nothing in the `3.0.0` breaking changes suggests anything we're using has changed.

alexlafroscia:
Closes #351

### DIFF
--- a/tests/unit/file-test.js
+++ b/tests/unit/file-test.js
@@ -1,0 +1,32 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
+import { upload as uploadHandler } from 'ember-file-upload/mirage';
+import File from 'ember-file-upload/file';
+
+module('Unit | File', function(hooks) {
+  setupTest(hooks);
+  setupMirage(hooks);
+
+  test('it can upload without a `queue`', async function (assert) {
+    assert.expect(1);
+
+    this.server.post('/image', uploadHandler((schema, request) => {
+      assert.deepEqual(
+        request.requestBody.file,
+        {
+          name: 'blob',
+          size: 12,
+          type: 'text',
+          url: 'data:text;base64,TXkgVGVzdCBGaWxl',
+          extension: undefined
+        },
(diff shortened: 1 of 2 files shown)


tests/unit/file-test.js
```diff
@@ -0,0 +1,32 @@
...
+import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
+import { upload as uploadHandler } from 'ember-file-upload/mirage';
+import File from 'ember-file-upload/file';
+
+module('Unit | File', function(hooks) {
+  setupTest(hooks);
+  setupMirage(hooks);
+
+  test('it can upload without a `queue`', async function (assert) {
+    assert.expect(1);
```
jelhan, line 12: Personally I prefer `assert.step()` and [`assert.verifyStep()`](https://api.qunitjs.com/assert/verifySteps) as it's easier to maintain when additional assertions are added. But that's more a question of coding style. `assert.expect()` works as well.

alexlafroscia:
Is it OK if I merge this? I’d rather an active maintainer actually does so but I have the right permissions to merge myself if no one has a problem with that.

alexlafroscia:
Thanks @jelhan !

alexlafroscia:
If it's possible to get a release cut with this bug-fix in it, I would really appreciate it! 😀

jelhan:
> If it's possible to get a release cut with this bug-fix in it, I would really appreciate it! grinning

Released 3.0.4 with this fix. Sorry for the delay.

alexlafroscia:
No problem -- thanks!

Issue #127, opened 2018-04-04
Validating based on extension during dragging is unreliable 


lukemelia:
Unfortunately, PR #121 does not copy out the file name, which is needed to validate based on extension. When I set out to add this, I discovered that some browsers (Chrome, at least) do not make the file name available while dragging. Therefore, it appears to be impossible to reliably validate during drag based on file extensions. Additionally, since some files (e.g., .xls on my Mac) have no mime type set, it appears to be impossible to rely on mime type exclusively except for very standard file types. This leaves us in a tough spot.

tim-evans:
Yeah, I believe this is a security issue, which makes it frustrating for us trying to make a better experience for our users. I'm open to suggestions for this, but I think the best approach will be to remove validation on drag & drop, since it's unreliable.

lukemelia:
I tend to agree @tim-evans. The workaround I just implemented (we'll see how it holds up) is that if my `accept` prop includes non-standard mime types, I have a CP that indicates that validation is not reliable. So while dragging, my dropzone subclass has three potential states: valid, invalid, and "can't reliably say".

tim-evans:
😂 Well, that sounds just about right.

PR #376, opened 2020-07-15
fix "always invalid" file on drag-drop event

mansona:
This is a first attempt to fix https://github.com/adopted-ember-addons/ember-file-upload/issues/127 in the most naive way possible 😂 

I would have implemented this in somewhere that is more specific to the dropzone component but it looks like `data-transfer` is only ever used once, and that is in the dropzone component.

This is mostly to open a discussion rather than request that this be merged, please let me know if we think there is a way forward on this issue and I can move the PR in that direction 👍


jelhan:
Hello Chris. Thanks for working on this one. I was wondering in which situations the file name is available and in which ones not. #127 mentions that it depends on the browser. Did you had a chance to research that one?

I would consider removing the validation of files added through drag and drop at all depending on browser support. That's why I'm asking.

mansona:
Hey @jelhan sorry for the delay, we have been working some more on this issue because it's slightly more complex than I originally thought 🙈 

We essentially lost the ability to notify the user that any file that was dropped on the dropzone was invalid because of the way that the validity actually only comes into play when it's "active" and since we can't get the file name during drag the API doesn't really make sense any more 😫 

We have extended it a bit further to get a set of `invalidFiles` that is exposed via the yielded API, that way we can loop over the invalid files after an upload and display this information to the end user 👍 

Also to answer your question, it is now described in the official spec that you cannot get access to the file information on the dragStart event: https://html.spec.whatwg.org/multipage/dnd.html#security-risks-in-the-drag-and-drop-model

Here is the relevant part: 

> User agents must not make the data added to the DataTransfer object during the dragstart event available to scripts until the drop event

mansona:
If you're ok with this direction we can start to discuss the more detailed aspects of this PR and I can write tests/docs or whatever is needed to get this landed. I just didn't want to do too much work before we agreed on the approach 👍 

For the record we have an implementation working with this now that can have multiple files dropped on it, the valid ones are uploaded and we can show the files that weren't uploaded as errors in place 🎉

mansona:
So, just to fill you in on the story so far. I think we've fixed the tests, but it looks like we might have stumbled on a strange use case that is not very well defined 🤔  

Essentially I notice that there are no tests that cover the case where `accept` has been set to **both** mime types and extensions. E.g. something like 

```
accept="image/*,.pdf"
```

The way that the code is implemented would suggest that you can mix and match these types of specifying what it accepts but we currently don't have any tests that cover this scenario.

It also means that we don't need to do the "hack" to assume the filename matches the extensions in the case of dragevent if you have not specified any extensions in your `accept` argument. This is what I did to fix the tests 👍 

Let me know if you want any more information about this!

st-h:
I can confirm that some browser do not provide much information while the drag is active. I have been running into this myself quite a while ago. (I wanted to show a count of how many of the dragged files are audio files). Turned out, this is impossible to implement.

Another thought - I don't want to mess with this PR in any way, though (this is very likely unrelated). Quite a while ago alon and I had a brief chat on discord, because he noticed that I was publishing a related addon: https://github.com/st-h/ember-file-drop-zone (which just provides a dropzone for files)

While trying to not duplicate ember addons, we came to the conclusion that it might be a good idea that ember-file-upload could use the drag and drop functionality that ember-file-drop-zone provides, instead providing its own implementation. However, so far I haven't had into time to further look into this.

jelhan:
To be honest I'm still struggling to understand the problem.

I got from the quoted specification that we don't have enough information available to validate on `dragStart`. But we should have that information - especially the file name - available on `dragEnd`, isn't it? Any reason we must validate on `dragStart` and can not wait until `dragEnd` event is fired? Or did I missed something?

mansona:
@jelhan yes there is a reason, especially with the current architecture 😄 

The issue is the interplay between `active` and `valid`  on the Dropzone object (or rather the dropzone object that is being yielded to the `<FileDropzone />` component: https://adopted-ember-addons.github.io/ember-file-upload/docs/api/components/file-dropzone

The dropzone is only considered `active` between `dragStart` and `dragEnd`, so once you have dropped the files (`dragEnd`) your dropzone is no longer active. In the current architecture the validation is happening while you're in the middle of a drag (before `dragEnd`) so you can't accurately determine if the files match the `accept` criteria because you just don't have the file names available to you at this time. 

This PR essentially defers this decision of if the files are valid until the `dragEnd` event (after you have dropped the files), uploads the valid files and keeps track of the invalid files in a way that can be communicated to the user if the developer wanted to 👍 

Does that help to explain the situation a bit more?

jelhan:
Thanks a lot for that explanation. Very helpful.

> In the current architecture the validation is happening while you're in the middle of a drag (before `dragEnd`) so you can't accurately determine if the files match the `accept` criteria because you just don't have the file names available to you at this time.

Do I get it right that this architecture is broken by design? The specification that you have quoted states that the required data *must not* be available at this point of time. So validation based on file name or file type can't work. If that's right, I don't have any objections removing the related code.

> This PR essentially defers this decision of if the files are valid until the `dragEnd` event (after you have dropped the files), uploads the valid files and keeps track of the invalid files in a way that can be communicated to the user if the developer wanted to :+1:

I have some questions and concerns regarding that API:

- Do I get it right that valid files would always be uploaded even if they were dropped together with invalid one? Would this support not uploading any of the dropped files if at least one of it was invalid?
- What user interaction resets the validation state? Are the validation errors shown until the user drops other files? Or is it cleared on `dragStart` / `activate` already? Does a `change` even on an embedded `<FileUpload>` resets it? Can the developer trigger a reset manually? By working on form components of Ember Bootstrap I learned that visibility and timing of validation states is a difficult topic and the requirements are different from project to project.
- Is there anything from the current API that prevents consuming applications to validate the input themselves?

To be honest I'm tending towards dropping validation support at all:

1. It's hard to get it right for most of the use cases if you take visibility, reset and timing questions into account. I think this is better served in user-land therefore. Developer should be able to use the same validation libraries as for other inputs (e.g. ember-changeset-validations or ember-cp-validations).
2. Files uploaded through `<FileUpload>` compontent (and native `<input type="file">`) are not validated. The `accept` argument / attribute does not provide validation. It's more like a hint to the user which files should be uploaded. A user can upload other files just by changing the filter on the file selection dialog persented by the browser. I don't see good reasons why `<FileDropzone>` should provide validation but `<FileUpload>` does not. I'm not aware that any one complained about missing validation on `<FileUpload>`.

mansona:
> Do I get it right that this architecture is broken by design? 

Yes the current architecture is broken by design 👍  (at least for the drag and drop implementation)

> validation based on file name or file type can't work.

only in the drag and drop implementation and again this only applies to cases where you are trying to do validation based on **filename**. I have a feeling that mime types are still working 🤔  but I haven't checked this in my example implementation

> * Do I get it right that valid files would always be uploaded even if they were dropped together with invalid one? Would this support not uploading any of the dropped files if at least one of it was invalid?

So... not quite... because the system is implemented as a kind of filter system we **do** have a situation where the valid files will be uploaded and the invalid files will not be uploaded in a drag and drop situation. Before my change the only indication that you would get is that the dropzone `valid` value would change to false in this situation **during** the drag phase and there would be no other indication after the drop.

There is a bit of a complicated situation here because the `addon/system/data-transfer.js` _class_ is being used to validate files during the drag event and then again just before uploading. And another thing that complicates the situation is that it is effectively implemented as a glorified filter, which makes it hard to reason about (and why I needed to add the `invalidFiles` implementation to be able to inspect the files that were filtered out). My implementation allows you to list the files that were invalid after a drop event so that you can give feedback to the user.

> * What user interaction resets the validation state? Are the validation errors shown until the user drops other files? Or is it cleared on `dragStart` / `activate` already? 

Right now the error will go away when you drag a new file on top of the dropzone

> Does a `change` even on an embedded `<FileUpload>` resets it? 

So... I'm not 100% sure exactly what you mean by this but I just tested clicking the dropzone and manually uploading a file (which interacts with the `<FileUpload />` rather than the dropzone) and that does **not** reset the error 🙈 

> Can the developer trigger a reset manually?

Not currently 🤔  we would need to expose a `resetError()` function in the yield if we wanted to have this functionality

> * Is there anything from the current API that prevents consuming applications to validate the input themselves?

This is really one of the hardest parts of the problem. When you pass `accepts` to either the Dropzone of FileUpload it sets that value on the **shared** queue object. This means that we can't set the `accepts=""` attribute on the `<input>` without initiating the full validation support fo the whole system. I would have been happy to implement a custom validation system in my user-land code but I cannot live with the trade-off of not being able to have the right `accepts=""` attribute on the `<input>`

> To be honest I'm tending towards dropping validation support at all

I think that would be a little heavy handed but if you think that is ultimately the right solution then it should be ok 👍 

> 1. It's hard to get it right for most of the use cases if you take visibility, reset and timing questions into account. I think this is better served in user-land therefore. Developer should be able to use the same validation libraries as for other inputs (e.g. ember-changeset-validations or ember-cp-validations).

While I think that we should be able to defer this to the user-land code (most probably in the `onfileadd` callback) but I am pretty sure that this is not something that we can rely on ember-cp-validations or ember-changeset-validations for 🤔  I have been doing a lot of work with ember-cp-validations lately and I have no idea how you might integrate it into a system like this 🙃  and I wouldn't want the task of trying to do that, especially in a modern Octane world.

> 2. Files uploaded through `<FileUpload>` compontent (and native `<input type="file">`) are not validated. 

That's not true 🤔  The files do in fact go through the same file path but it's quite difficult to test because of the fact that the `accepts=""` attribute prevents me from picking an invalid file.

> The `accept` argument / attribute does not provide validation. It's more like a hint to the user which files should be uploaded. A user can upload other files just by changing the filter on the file selection dialog persented by the browser. 

eh... I'm not entirely sure this is true 🤔  at least on a Mac this is 100% **not true**. There is no way for me to change the type that I am selecting from and I'm 100% restricted to what I have defined in `accept=""`. 

> I don't see good reasons why `<FileDropzone>` should provide validation but `<FileUpload>` does not. I'm not aware that any one complained about missing validation on `<FileUpload>`.

Even if you were able to pick something that was not following the `accept=""` attribute on the `<input>` the files do go through the same code-path so they **are** validated 👍

jelhan:
> > The `accept` argument / attribute does not provide validation. It's more like a hint to the user which files should be uploaded. A user can upload other files just by changing the filter on the file selection dialog persented by the browser.
> 
> eh... I'm not entirely sure this is true  at least on a Mac this is 100% **not true**. There is no way for me to change the type that I am selecting from and I'm 100% restricted to what I have defined in `accept=""`.

This seems to be related to browser and/or operation system. I just tried again with Chromium running on Ubuntu. It only shows images, which are matching the `accept` attribute first. But I'm able to switch the filter to all files. I can select any file afterwards.

> > I don't see good reasons why `<FileDropzone>` should provide validation but `<FileUpload>` does not. I'm not aware that any one complained about missing validation on `<FileUpload>`.
> 
> Even if you were able to pick something that was not following the `accept=""` attribute on the `<input>` the files do go through the same code-path so they **are** validated

I just tried it again on the [demo application](https://adopted-ember-addons.github.io/ember-file-upload/). I created a text file `examples.txt`. It should not be allowed by the `accept` argument, which is `image/*,video/*,audio/*`. First I tried to upload it using drag and drop functionality. It was not shown in the list. So I assume it was rejected by validation. Afterwards I tried to upload it through the `<input file="file">`. I was able to select it in Chromuim as explained above. `examples.txt` was shown as being added in the list after selecting it. Did I missed something?

st-h:
> This seems to be related to browser and/or operation system. I just tried again with Chromium running on Ubuntu. It only shows images, which are matching the accept attribute first. But I'm able to switch the filter to all files. I can select any file afterwards.

I am pretty sure that safari changed this somewhen to be no longer configurable by the user. In Chrome on Mac there is an options button, which then lets you remove the restriction to certain files. No such thing with safari. And then to make it reliable, there is always that issue, if just checks by file suffix or if it checks the actual file type.

mansona:
@jelhan the examples you are giving are using the mime types, I am only talking about the file extensions when I'm discussing this issue. I recommend setting accept to be `.jpg` and see what happens in this scenario 🤔

jelhan:
> @jelhan the examples you are giving are using the mime types, I am only talking about the file extensions when I'm discussing this issue. I recommend setting accept to be `.jpg` and see what happens in this scenario :thinking:

I tried once again.

1. I changed the `accept` argument to `.jpg` in dummy app:
    ```patch
    diff --git a/tests/dummy/app/templates/index.hbs b/tests/dummy/app/templates/index.hbs
    index a056462..35efb5c 100644
    --- a/tests/dummy/app/templates/index.hbs
    +++ b/tests/dummy/app/templates/index.hbs
    @@ -45,7 +45,7 @@
                   for="upload-photo"
                   multiple=true
                   onfileadd=(action "uploadProof")
    -              accept="image/*,video/*,audio/*"
    +              accept=".jpg"
                 }}
                   <a>select them</a>
                 {{/file-upload}}
    ```
2. I created an `example.jpg` and `example.png`. 
3. I tried to upload both using drag and drop. The `example.png` was not shown in the list of uploaded files.
4. I tried to upload both using the `<FileUpload>` component. I selected the `example.png` even if it was not matching the accept attribute as discussed above. Both files were shown in the list of uploaded files.

knownasilya:
I've found that on drag the file.name isn't set in the DataTransfer items (https://github.com/adopted-ember-addons/ember-file-upload/blob/master/addon/system/drag-listener.js#L111) but then expected here https://github.com/adopted-ember-addons/ember-file-upload/blob/master/addon/system/data-transfer.js#L99

Issue #377, opened 2020-07-20
import of mirage upload handler causes fastboot error


gibso:
I am using this addon in a large ember-application with fastboot and mirage. So far, it works like a charm. But as soon as i want to import the upload helper in my mirage config ...

```javascript
// mirage/config.js

import { upload } from 'ember-file-upload/mirage';
``` 
... Fastboot throws an error when i call any route in my browser. When i turn off fastboot everything works fine. Am i missing something?

Thank you very much in advance!

```
ReferenceError: FormData is not defined
    at Module.callback (/tmp/broccoli-1COZmebBW7o5T/out-684-broccoli_merge_trees/assets/addon-tree-output/ember-file-upload/mirage/shim.js:11:1)
    at Module.exports (/tmp/broccoli-1COZmebBW7o5T/out-684-broccoli_merge_trees/assets/vendor/loader/loader.js:106:1)
    at Module._reify (/tmp/broccoli-1COZmebBW7o5T/out-684-broccoli_merge_trees/assets/vendor/loader/loader.js:143:1)
    at Module.reify (/tmp/broccoli-1COZmebBW7o5T/out-684-broccoli_merge_trees/assets/vendor/loader/loader.js:130:1)
    at Module.exports (/tmp/broccoli-1COZmebBW7o5T/out-684-broccoli_merge_trees/assets/vendor/loader/loader.js:104:1)
    at Module._reify (/tmp/broccoli-1COZmebBW7o5T/out-684-broccoli_merge_trees/assets/vendor/loader/loader.js:143:1)
    at Module.reify (/tmp/broccoli-1COZmebBW7o5T/out-684-broccoli_merge_trees/assets/vendor/loader/loader.js:130:1)
    at Module.exports (/tmp/broccoli-1COZmebBW7o5T/out-684-broccoli_merge_trees/assets/vendor/loader/loader.js:104:1)
    at Module._reify (/tmp/broccoli-1COZmebBW7o5T/out-684-broccoli_merge_trees/assets/vendor/loader/loader.js:143:1)
    at Module.reify (/tmp/broccoli-1COZmebBW7o5T/out-684-broccoli_merge_trees/assets/vendor/loader/loader.js:130:1)
    at Module.exports (/tmp/broccoli-1COZmebBW7o5T/out-684-broccoli_merge_trees/assets/vendor/loader/loader.js:104:1)
[...]

PR #378, opened 2020-07-20
remove mirage FormData shim to fix fastboot error

gibso:
this fixes issue #377 

the import of the mirage upload handler causes a fastboot error. Removing this shim solves the problem.


gilest:
Browser support for `FormData.get()`

Screen Shot 2020-07-21 at 10 19 18 AM

jelhan:
Thanks a lot @gilest for including the browser support matrix. Seems like we don't need this shim anymore since we dropped IE11 support or did I missed something? If so, this should be good to merge. Thanks a lot @gibso.

gilest:
> Seems like we don't need this shim anymore since we dropped IE11 support or did I missed something?
You didn't miss anything, just wanted to make sure we don't break anything for browsers we do support.

PR #396, opened 2020-09-11
docs: add link to accept argument docs

knownasilya:
Just to clarify so users don't have to search the code to see how it should be formatted


jelhan (approved):
Thanks a lot!

PR #427, opened 2021-04-15
Move mirage handlers from /addon to /addon-test-support

### LINKED ISSUE #425: Move `mirage` folder outside of `addon` folder
By default everything included in the `addon` folder ships. Since Mirage is a mocking tool, related code should not be included in the final build (at least not automagically).

Screenshot 2021-03-30 at 13 20 29

Maybe the same can be accomplished by moving it to the `addon-test-support` folder, but haven't tested this myself.

gilest:
Tests pass locally with these changes 🤷 

Would this a breaking change since the import path changes?

What labels should I assign this PR? `breaking`?

Fixes #425


jelhan:
I'm not sure if this change is appropriate. Mirage may be used in builds in which tests are not included. For such builds mirage handlers would not be available.

I think we must include mirage handlers in the build depending if mirage itself us included. I guess Ember CLI Mirage's build hooks could provide some inspiration how to do that.

bertdeblock:
@jelhan I know that `ember-cli-mirage-graphql` does this as well. [It replicates the behaviour from ember-cli-mirage](https://github.com/kloeckner-i/ember-cli-mirage-graphql/blob/master/index.js#L37-L49). Would be nice if `ember-cli-mirage` could provide an API or something for this, so the actual logic remains there.

gilest:
Closing as this would remove documented usage – use of mirage handlers in development/production builds. We're exploring other options in  #425

PR #428, opened 2021-04-15
Add embroider scenarios to ember-try

gilest:
Currently allowing both builds to fail.

We should probably migrate to GH actions...


gilest:
Scenarios were added in #430. Closing

PR #429, opened 2021-04-15
Add Github Actions workflow

gilest:


jelhan:
Thanks a lot for starting to work on this.

> Haven't configured the Code Climate coverage stuff – do we want to port this?

If we decide to _not_ port it, we could use [create-github-actions-setup-for-ember-addon](https://github.com/jelhan/create-github-actions-setup-for-ember-addon). Missing support for Code Climate and similar customization of CI pipeline is the only reason I haven't migrated this addon yet. Using the script to create (and update) GitHub Actions workflows is also [recommended by program guidelines for adopted ember addons](https://github.com/adopted-ember-addons/program-guidelines#continuous-integration).

gilest:
@jelhan

I'm struggling to fix this one issue.

> There's a test failure in the mirage-handler test that checks for errors emitted from uploading an invalid audio file. It's expected that an error from within a promise is elevated to ember's error handler – but this doesn't happen in some builds.

> Doesn't occur in my local test runs. Can sort-of reproduce by adding a small timeout or a debugger [here](https://github.com/adopted-ember-addons/ember-file-upload/blob/master/addon/mirage/utils.js#L118).

I've tried:

- Also listening to `RSVP.on('error')`
- Upgrading to `ember-qunit` 5
- Many, many dependency changes (see latest ember try builds). The test failure is not consistent with specific packages as far as I can tell
- Regenrating `yarn.lock`

jelhan:
> I'm struggling to fix this one issue.
> 
> > There's a test failure in the mirage-handler test that checks for errors emitted from uploading an invalid audio file. It's expected that an error from within a promise is elevated to ember's error handler – but this doesn't happen in some builds.
> 
> > Doesn't occur in my local test runs. Can sort-of reproduce by adding a small timeout or a debugger [here](https://github.com/adopted-ember-addons/ember-file-upload/blob/master/addon/mirage/utils.js#L118).

Tests, which fail on CI but not locally, are the worst. So challenging to debug...

This is the test, which is failing: https://github.com/adopted-ember-addons/ember-file-upload/blob/7f8d7c4f2b12a6971a7fff249cede444ee66efc2/tests/integration/components/mirage-handler-test.js#L57-L103

The same test code is run for three scenarios: audio, image and video. Only audio is failing.

Audio and video are handled very similar in upload handler provided for mirage: https://github.com/adopted-ember-addons/ember-file-upload/blob/7f8d7c4f2b12a6971a7fff249cede444ee66efc2/addon/mirage/utils.js#L76-L126 One creates a `<audio>` the other a `<video>` element. The video function reads `videoWidth` and `videoHeight` attributes additionally. I would be surprised if that small difference causes one to pass and the other to fail.

I was wondering if the CI is flickering or fails always for the same scenarios. I will restart several times and note which scenarios are passing and which are failing.

| scenario | first run                                       | second run | third run |
| ----------- | ------------------------------------------ | --------------- | ----------- |
| release  | :negative_squared_cross_mark: |
| 3.24.3    | :heavy_check_mark:                   |
| 3.20.6    | :negative_squared_cross_mark: |
| 3.16.10 | :negative_squared_cross_mark:  |
| 3.12.4   | :negative_squared_cross_mark: |
| 3.8.3     | :heavy_check_mark:                   |

jelhan:
@gilest I'm wondering if #430 might fix the CI. At least TravisCI passed for that feature branch but is failing on `master`. Maybe we land #430 first and migrate to GitHub Actions afterwards?

gilest:
> @gilest I'm wondering if #430 might fix the CI. At least TravisCI passed for that feature branch but is failing on `master`. Maybe we land #430 first and migrate to GitHub Actions afterwards?

Yes, let's try that. We need to keep moving anyway!

jelhan:
@gilest I merged #430. Would be great if you could rebase / merge latest master in.

gilest:
@jelhan rebased and re-ran `create-github-actions-setup-for-ember-addon`

Seems like that one spec is still unreliable under GitHub Actions

jelhan:
> Seems like that one spec is still unreliable under GitHub Actions

I would tend to skip that test for now. It only covers mirage test helpers. And this is eating up too much of our limited time...

gilest:
@jelhan Ok I've skipped that test and we have fields of green now ✅ 

Only thing remaining is to migrate the docs deployment.

I've configured a [basic workflow](https://github.com/adopted-ember-addons/ember-file-upload/pull/429/files#diff-28802fbf11c83a2eee09623fb192785e7ca92a3f40602a517c011b947a1822d3).

Just need a deployment key to be generated and added to GitHub secrets. I didn't know who has the permissions to do this so I've asked for help on Discord.

gilest:
@jelhan I believe this is ready now 🙌 

Have tested a docs deploy with the new configuration and it worked as expected.

Please review 👀

tests/integration/components/mirage-handler-test.js
```diff
@@ -76,7 +76,7 @@ module('Integration | Component | mirage-handler', function (hooks) {
     }),
   };
   for (let [type, file] of Object.entries(SCENARIOS)) {
-    test(`upload handler throws if invalid ${type} is provided`, async function (assert) {
+    skip(`upload handler throws if invalid ${type} is provided`, async function (assert) {
```
jelhan, line 79: Wondering if we should only skip the test for `type === 'audio'`. The other two scenarios were stable, weren't they? Something like:

```js
     (type === 'audio' ? skip : test)(`upload handler throws if invalid ${type} is provided`, async function (assert) {
```

gilest: Good point. Will add this now

gilest: I had to skip both `audio` and `video` types as they both fail in the same way

jelhan: Oh. I missed that. Having both failing in the same way actually makes it less confusing. 😄

jelhan:
The deployment of docs failed when triggered by v4.0.0 release: https://github.com/adopted-ember-addons/ember-file-upload/runs/2873915337 @gilest Would be great if you could have a look what is going on.

gilest:
@jelhan Ah yes – looks like the branch and tag deploys raced. Re-ran the workflow and it deployed fine.

Permanent fix in #481

PR #431, opened 2021-04-17
Pin incompatible dependencies

gilest:
As sub-dependencies out in the wild have been updated they have caused the "Floating Dependencies" builds to fail.

- ember-data: Too permissive. The latest version 3.26 was being installed. Incompatible with the existing addon
- whatwg-fetch: Incompatible version. Can be cleared up when we next upgrade addon-docs. (https://github.com/ember-cli/ember-fetch/issues/547)

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,9 @@
     "postversion": "git push && git push --tags"
   },
   "resolutions": {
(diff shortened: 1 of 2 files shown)


jelhan (approved):
Looks good to me. Scenarios for beta and canary releases of Ember are still failing. But I guess that's not related.

Issue #449, opened 2021-05-27
Fix this-property-fallback deprecation


aoumiri:
`this-property-fallback` deprecation warnings in `components/file-upload/template.hbs` and `components/file-dropzone/template.hbs` may lead to a lot of noise in CI pipelines for the applications that uses those components intensively.

A PR to solve this issue will come right after 🌟

gilest:
Closed by #450

PR #450, opened 2021-05-27
Fix this-property-fallback deprecation

aoumiri:
This PR solves #449

### DIFF
--- a/addon/components/file-dropzone/template.hbs
+++ b/addon/components/file-dropzone/template.hbs
@@ -1,1 +1,1 @@
-{{yield (hash supported=supported active=active valid=valid) queue}}
+{{yield (hash supported=this.supported active=this.active valid=this.valid) this.queue}}
--- a/addon/components/file-upload/template.hbs
+++ b/addon/components/file-upload/template.hbs
@@ -1,12 +1,12 @@
 <input
-  id={{for}}
+  id={{this.for}}
   type="file"
-  accept={{accept}}
-  capture={{capture}}
-  multiple={{multiple}}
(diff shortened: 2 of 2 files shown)


jelhan (approved):
Thanks a lot for picking this one up.

In many cases `{{this.foo}}` is the same as `{{@foo}}`. But as the arguments are defined as properties on the component class I agree that we should reference it in the template. Everything else would be consuming. We should refactor to `{{@foo}}` when migrating to `@glimmer/component`.

jelhan:
> > In many cases `{{this.foo}}` is the same as `{{@foo}}`. But as the arguments are defined as properties on the component class I agree that we should reference it in the template. Everything else would be consuming. We should refactor to `{{@foo}}` when migrating to `@glimmer/component`.
> 
> I was wondering if I should change that too, but I figured it would be a bit more adventurous refactoring. I would be happy to do it in another PR though smile

A pull request to refactor the components to `@glimmer/component` is very welcome.

jelhan:
@gilest This change seems safe to me. Is it okay for you to merge even though CI is failing current and release as the last version of v3?

gilest:
@jelhan yes seems fine to me too. Go ahead and merge

gilest (approved):

aoumiri:
Thanks everyone for your help!

gilest:
Thank you @aoumiri

jelhan:
Published as [`v3.0.6`](https://github.com/adopted-ember-addons/ember-file-upload/releases/tag/v3.0.6).

PR #475, opened 2021-06-20
Ignore progress events where the total file size is reported as 0 bytes

### LINKED ISSUE #344: File progress is sometimes NaN
This appears to happen because total is 0 and you can't divide by 0.

gilest:
Have tested this manually using the docs site. Wasn't able to reproduce a `NaN` value.

### Before

- Drop a file
- Progresses from 0 => 100%
- Progress set back to 0% for a moment
- Finishes

### After

- Drop a file
- Progresses from 0 => 100%
- Progress stays at 100% for a moment
- Finishes

Fixes #344

### DIFF
--- a/addon/file.js
+++ b/addon/file.js
@@ -74,11 +74,11 @@ function upload(file, url, opts, uploadFn) {
   }
 
   request.onprogress = function (evt) {
-    if (evt.lengthComputable) {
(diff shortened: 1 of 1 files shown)


jelhan (approved):
Looks good to me. Thanks for picking this old bug up!

PR #476, opened 2021-06-20
Renamed all events to camelCase

gilest:
Includes `deprecatingAlias` support for old event names until 5.0.0

Added component tests to cover both new and deprecated event names. The duplicated tests are quick and dirty but they will all helpfully fail as soon as the alias is removed.

See #328 for previous discussion

### DIFF
--- a/addon/components/base-component.js
+++ b/addon/components/base-component.js
@@ -9,8 +9,8 @@ export default Component.extend({
   didReceiveAttrs() {
     this._super(...arguments);
     if (this.queue) {
-      let { accept, multiple, disabled, onfileadd } = this;
-      setProperties(this.queue, { accept, multiple, disabled, onfileadd });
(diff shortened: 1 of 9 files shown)


jelhan (approved):
Looks good to me. 👍

Wondering if we should still include the deprecated aliases in API docs. We would need to duplicate the existing comment blocks for documentation to do so. And add a `@deprecated`.

I tend towards skipping them. People using them are pointed to correct documentation by the deprecation. And for others it may only cause confusion.

gilest:
> Wondering if we should still include the deprecated aliases in API docs. We would need to duplicate the existing comment blocks for documentation to do so. And add a `@deprecated`.

Hmm yes I hadn't considered this. Thanks for pointing it out.

Let's keep this in mind – if issues start getting opened we could always update the docs. 

> I tend towards skipping them.

Agree, I think it's clear enough as it is for now.

PR #481, opened 2021-06-21
Add concurrency to deploy workflow

gilest:
Should cause all deploy jobs to be queued and solve issues with deploys racing each other.

We saw an issue recently when we tagged a release where the `tag` and `master` deploys raced each other, causing the slower deployment to fail.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,5 +1,7 @@
 name: Deploy
 
+concurrency: deploy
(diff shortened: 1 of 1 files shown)


jelhan (approved):
Thanks a lot for the nice description of what has caused the bug and how it is solved. To be honest I'm not that used to GitHub Actions yet. Great opportunity to learn more. :smile:

Issue #316, opened 2020-02-28
Improve public API


gilest:
Since @jelhan brought up significant API changes in #306 . I thought I'd post some of my "RFC" notes about the public APIs.

Interested to hear thoughts.

```handlebars
{{!-- Current 2.7 API --}}

{{#file-dropzone name="photos" as |dropzone queue|}}
  {{#if dropzone.active}}
    {{#if dropzone.valid}}
      Drop to upload
    {{else}}
      Invalid
    {{/if}}
  {{else if queue.files.length}}
    Uploading {{queue.files.length}} files. ({{queue.progress}}%)
  {{else}}
    <h4>Upload Images</h4>
    <p>
      {{#if dropzone.supported}}
        Drag and drop images onto this area to upload them or
      {{/if}}
      {{#file-upload name="photos"
                     accept="image/*"
                     multiple=true
                     onfileadd=(route-action "uploadImage")}}
        <a id="upload-image" tabindex=0>Add an Image.</a>
      {{/file-upload}}
    </p>
  {{/if}}
{{/file-dropzone}}

{{!--
  3.0/4.0? API RFC

  Contextual component

  * Attach all queue callbacks via the contextual/provider component
  * Rename FileUpload component to FileSelect, because FileUpload is too generic
  * FileSelect and FileDropzone _require_ a queue to be passed in, no more name attribute
  * Dropzone and FileSelect can only add to the queue, and provide callbacks for their own actions
  * Disadvantage: the most simple usage is a little more verbose
  * Advantage: maybe easier to use, less magic, more flexible
 --}}

<FileUploadQueue @name="photos" @onFileAdd={{action this.uploadImage}} as |queue|>
  <FileDropzone @queue={{queue}} @onDrop={{action this.filesSelected}} as |dropzone|>
    {{#if dropzone.active}}
      {{#if dropzone.valid}}
        Drop to upload
      {{else}}
        Invalid
      {{/if}}
    {{else if queue.files.length}}
      Uploading {{queue.files.length}} files. ({{queue.progress}}%)
    {{else}}
      <h4>Upload Images</h4>
      <p>
        {{#if dropzone.supported}}
          Drag and drop images onto this area to upload them or
        {{/if}}
        <FileSelect @queue={{queue}}
                    @for="upload-photo"
                    @accept="image/*"
                    @multiple={{true}}
                    @onSelected={{action this.filesSelected}}>
          <a tabindex="0">Add an Image.</a>
[...]

PR #514, opened 2021-09-01
Upgrade to glimmer components and native classes

gilest:
Migration to Octane edition. Deals with all legacy ember objects, components, mixins and general tech debt.

Attempting to support the existing public API so that we can release this as a `v4.x.0` minor. Possibly with a beta first to make sure there are no regressions.

### Queue instances and the FileQueue service
Reworked this relationship significantly.

Previously File objects were added to the FileQueue service as well as the Queue instances themselves. They shared some behaviour for this via the `WithFiles` mixin.

Now the FileQueue service declaratively accesses files within all Queue instances to provide its public API. All queue flushing has been removed from the FileQueue service – it's the responsibility of the Queue instances only.

### Queue flushing

After refactoring away from observers (#307) we added manual calls to `Queue.flush()` after any internal file state changes.

This required that we warn users they may need to flush manually if they change the file state property themselves:

> Implementers should call flush() on the file's queue after mutating this property.

I've refactored this by using a setter to call `Queue.flush()` on any file state changes and removed this piece of documentation.

### Queue instances and Components

When `FileUpload`/`FileDropzone` call `didRecieveAttrs` they directly set some properties on their queue.

I've refactored this to an `update-queue` modifier.

```hbs
  {{update-queue
    this.queue
    accept=@accept
    multiple=@multiple
    disabled=@disabled
    onFileAdd=this.onFileAdd
  }}
```

When testing this I noticed that it's an unsafe operation – the queue only has a single property for each of these and there's no guarantee about the order in which properties might overwrite each other. (For example when rendering a dropzone and a file select for the same queue).

I'm comfortable releasing as-is for now (believe it's the same as existing functionality) – but we should reconsider this API in the future.

### DragListener

I've also refactored this into a modifier.

```hbs
  {{drag-listener
    dragenter=this.didEnterDropzone
    dragleave=this.didLeaveDropzone
    dragover=this.didDragOver
    drop=this.didDrop
  }}
```

This required me to adjust the `DragListener` class to work with an element rather than a selector.

### Test waiters

Migrated from old `registerWaiter` to the new `buildWaiter` from `@ember/test-waiters`.

### Component DOM

Almost identical to before. Since `FileDropzone` renders a wrapper div I added `...attributes` to it.

`FileUpload` is a bit more complex with the label/input relationship and existing attribute args so I left it the same as before.

I also removed an assertion about the wrapper tag of `FileUpload` as it's no longer possible to pass in `tagName` and override it.

### Dependencies

Since I added modifiers I've added `ember-modifier` to dependencies.

Also added glimmer packages and removed `@babel/core` - not sure why it was ever included.

## Questions

- Have I set up the modifiers properly? They are only for internal use.
- Are there any dependencies I've failed to add?
- I've used optional chaining `?.` and nullish coalescing `??`. Is this safe? Do I need to configure something for this?
- I added a dir `addon/utils` is that safe? Will that work?
- Any properties I forgot to track?


addon/file.js
```diff
@@ -1,14 +1,12 @@
...
-import { DEBUG } from '@glimmer/env';
-
 import { assert } from '@ember/debug';
-import EmberObject, { set, computed } from '@ember/object';
 import FileReader from './system/file-reader';
 import HTTPRequest from './system/http-request';
 import RSVP from 'rsvp';
 import uuid from './system/uuid';
+import { tracked } from '@glimmer/tracking';
+import { buildWaiter } from 'ember-test-waiters';
```
bertdeblock, line 7: Should this be `import { buildWaiter } from '@ember/test-waiters';` now?

addon/services/file-queue.js
```diff
@@ -37,7 +35,12 @@ export default class FileQueueService extends Service.extend(WithFiles) {
     @type {File[]}
     @default []
    */
-  files = A([]);
+  get files() {
```
bertdeblock, line 38: Should this still return an Ember array?

package.json
```diff
@@ -47,13 +47,16 @@
...
+    "@glimmer/tracking": "^1.0.4",
     "ember-cli-babel": "^7.26.3",
-    "ember-cli-htmlbars": "^5.7.1"
+    "ember-cli-htmlbars": "^5.7.1",
+    "ember-modifier": "^2.1.2"
   },
   "devDependencies": {
     "@ember/optional-features": "^2.0.0",
     "@ember/test-helpers": "^2.2.5",
+    "@ember/test-waiters": "^3.0.0",
```
bertdeblock, line 59: Should be a normal dependency I think since it's used in `addon/file.js`?

package.json
```diff
@@ -47,13 +47,16 @@
...
     "@embroider/test-setup": "^0.44.0",
     "@glimmer/component": "^1.0.4",
```
bertdeblock, line 61: Can be removed here I think, same for `@glimmer/tracking` as they are now defined in `dependencies`.

addon/file.js
gilest: Good catch – thanks! Fixed in 1ee1569c34f3c51f1fbe0a4d5966e318a20c5aab

addon/services/file-queue.js
gilest: Yeah I think it should. Good call - fixed in 2e0c082c3fc906e0f2d916d587e3f2507ec9d5f0

package.json
gilest: Right – fixed in 1ee1569

package.json
gilest: Ah 😅 I swear I did this already. Fixed in 9c941ff0496853572bf942662d6df4c27dd3e2ed

Thanks for the review 🙇

addon/components/file-upload.hbs
```diff
@@ -0,0 +1,26 @@
+<label
```
bertdeblock, line 1: I think we should add `...attributes` here as well? I'm guessing people might have been using `@class` for additional styling. If we add `...attributes` they can switch to `class`.

gilest: Ah.. yes 😅  I'm a bit conflicted about what to do here. This has been discussed a bit already in #293 and #316.

Ideally this PR paves the way for a full breaking-change rebuild of the public API.

I thought I'd pay all the tech debt while we're deciding the specifics, and then we can release that as a minor much more quickly. It means we won't be blocking anyone upgrading to Ember 4 either.

@jelhan is kindly going to review this soon too, I'll wait to hear his thoughts.

bertdeblock: Woops, sorry! I haven't seen those discussions. Just thought I'd mention it just in case.

gilest: On balance I've reconsidered this and added `...attributes` to the <label> tag

PR #567, opened 2021-10-14
feat: conditionally exclude mirage handler from builds

gilest:
Continuation of #432

@bendemboski 

> Fix CSP for embroider builds by switching the source mapping method to not use eval()
> 
> Note that rather than disabling source maps, I switched to a non-eval form. It's a slow method than others, but it only affects development/testing of ember-file-upload itself, and its build is fast enough that the increased build time is pretty insignificant and, IMO, worthwhile to have sourcemaps.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -26,5 +26,10 @@ module.exports = function (defaults) {
         package: 'qunit',
       },
     ],
+    packagerOptions: {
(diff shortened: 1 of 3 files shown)


gilest:
Yeah I checked that but we already had the `forbidEval` config set. Appreciate you guys finding a way forward 🙇🏻

PR #570, opened 2021-10-15
Fix CSP for embroider builds

bendemboski:
Fix CSP for embroider builds by switching the source mapping method to not use eval()

@gilest please feel free to just copy these changes into your PR/rebase for a clean commit history, I don't need "credit" 😄 

Note that rather than disabling source maps, I switched to a non-eval form. It's a slow method than others, but it only affects development/testing of `ember-file-upload` itself, and its build is fast enough that the increased build time is pretty insignificant and, IMO, worthwhile to have sourcemaps.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -26,5 +26,10 @@ module.exports = function (defaults) {
         package: 'qunit',
       },
     ],
+    packagerOptions: {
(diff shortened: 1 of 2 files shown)


gilest:
Thank you @bendemboski 

Cherry picked and squashed into #567

PR #602, opened 2021-11-15
chore(deps): add @ember/legacy-built-in-components

gilest:
They are imported by ember-keyboard but longer provided by ember-source in v4.

Can remove this package once ember-keyboard v7 is finished.

Reference: adopted-ember-addons/ember-keyboard/issues/491


gilest:
Not possible as this addon uses different import paths.

Closed in favour of #603

PR #603, opened 2021-11-15
fix(ember 4 builds): exclude unused ember-keyboard initializer

gilest:
It was breaking ember 4 builds.

Is this an acceptable way to deal with that?

I need to read the ember cli docs to make sure this won't exclude this file for addon consumers who actually want it.

Ref: adopted-ember-addons/ember-keyboard/issues/491


gilest:
Resolved in #604 via adopted-ember-addons/ember-keyboard/pull/498

PR #620, opened 2021-12-14
Add events to file-queue helper set up TypeScript, drop ember-cli-addon-docs

gossi:
Based on discussion in #316 - here is what I did:

Modernizations:

- Use `ember-cli-typescript` (not entirely, but for the key parts)
- Dropped `Ember.A()` (on those files I touched with TS) in favor of `@tracked`
- Use `@ember/destroyables` (depending on support matrix, this requires to ship the polyfill)
- I think this also prepares for addon format v2

Changes:

- Mainly the `{{file-queue}}` helper, which can be beautifully used like so:

```hbs
{{#let (file-queue fileAdded=this.addFile fileRemoved=this.removeFile) as |queue|}}
  {{#each queue.files as |file|}}
    <span data-test-file>
    {{file.name}}
    </span>
    <button
      type="button"
      data-test-remove
      {{on "click" (fn queue.remove file)}}
    >
      x
    </button>
  {{/each}}

  <label>
    <input type="file" {{queue.selectFile filter=this.filter filesSelected=this.selectFiles}} hidden>
    Select File
  </label>
{{/let}}
```

- I kept original API of `<FileUpload>` component and used new helper under the hood, rewired appropriately
- `FileQueueService` can now work with `DEFAULT_QUEUE` name (instead of super-global queue)

Cutbacks:

- With usage of TS I had to disable `ember-cli-addon-docs` (maybe it is time to move, ie. field-guide or docfy?)
- Had to drop `@onSelect` (couldn't rewire to the new API - I don't see this as an issue, the API was never public)

Questions:

- I think, a lot of things can be marked deprecated for v5 and removed in v6
- I left quite some `@TODO` in the code, some things I marked as removal/deprecation candidates or questions I have

What I didn't do (yet):

- No API for the upload process, ie. `(file-queue uploadStarted=... upload=...)` or `Queue.upload()` - where does `upload=()` need to be attached? To `(file-queue)` or `Queue`?
- I haven't touch drag'n'drop at all. I event kept `Queue._addFiles()` - which is sooo ready to go 😅

-> ALL THAT by keeping the TestSuite green - overall was pretty happy to manage backwards compatibility (that went way better than I thought !!). I also checked with our app, by linking this and let the app testsuite run, green as well.


addon/services/file-queue.ts
```diff
@@ -0,0 +1,126 @@
...
+    return queue;
+  }
+
+  findOrCreate(name: QueueName): Queue {
+    return this.find(name) ?? this.create(name);
+  }
+
+  //
+  // @TODO
+  // Everything below this line should be deprecated ?
```
gilest, line 62: I guess this API is confusing since there are some shared properties with `Queue` right? Not sure how much they are used by addon consumers.

package.json
```diff
@@ -53,18 +55,43 @@
...
     "ember-cli-babel": "^7.26.6",
     "ember-cli-htmlbars": "^5.7.1",
-    "ember-modifier": "^2.1.2 || ^3.0.0"
+    "ember-cli-typescript": "^4.2.1",
+    "ember-modifier": "^2.1.2 || ^3.0.0",
+    "tracked-built-ins": "^2.0.0"
   },
   "devDependencies": {
     "@ember/optional-features": "^2.0.0",
+    "@ember/render-modifiers": "^2.0.0",
```
gilest, line 64: As mentioned above this would need to be in `dependencies`. I favour using a custom modifier anyway.

gilest (changes requested):

addon/components/file-upload.ts
```diff
@@ -0,0 +1,125 @@
...
+import { inject as service } from '@ember/service';
+import { action } from '@ember/object';
+
+import uuid from '../system/uuid';
+import File from '../file';
+import Queue from '../queue';
+import FileQueueService, { DEFAULT_QUEUE } from '../services/file-queue';
+
+interface FileUploadArgs {
+  name?: string;
```
gilest, line 11: I think we should drop or at least deprecate this in favour of `queue`.

gilest:
This is a great contribution – thank you @gossi ! Hugely appreciate you wading in here.

### TS and docs

> Use ember-cli-typescript (not entirely, but for the key parts)

Appreciate that TS is very nice and helpful – and I don't want to discourage you – but it's a bit presumptuous to include without at least discussing first. Would also prefer to make this change in a single PR rather than bundle it with an API update.

> With usage of TS I had to disable ember-cli-addon-docs (maybe it is time to move, ie. field-guide or docfy?)

Are you volunteering to take care of this? I don't want to release without a docsite and the current one hosts versioned docs back to 2.4.4 which we'd have to archive somehow.

### General responses

> Use @ember/destroyables (depending on support matrix, this requires to ship the polyfill)

Yes we'll need `ember-destroyable-polyfill` as we currently support `v3.16` and above.

> Had to drop @onSelect (couldn't rewire to the new API - I don't see this as an issue, the API was never public)

It was recently added and documented in #578. It has been published in betas but not an official release.

Looks like with this api `filter` takes care of filtering and `filesSelected` is available as a callback for file selection. Seems good 👍🏻 Someone will need to update the docs.

### Component API

> I think, a lot of things can be marked deprecated for v5 and removed in v6

True. I would be happy to more aggressively update `<FileUpload>`. At least by removing the attribute arguments and using `...attributes`. This would be a very easy upgrade to document for users.

I guess regardless I would recommend addon consumers use the `queue.selectFile` method over `<FileUpload>` anyway... The main outstanding issue we hear about is with DOM structure and inaccessibility of hidden file input.

> No API for the upload process, ie. (file-queue uploadStarted=... upload=...) or Queue.upload() - where does upload=() need to be attached? To (file-queue) or Queue?

Adding upload callbacks to the `file-queue` helper makes sense to me. This would need to be registered like the other callbacks.

> I haven't touch drag'n'drop at all. I event kept Queue._addFiles() - which is sooo ready to go 😅

This can mostly be handled internally since the API changes are happening at the queue level, yeah?

We will need to at least allow `@queue` to be passed in as an arg though. See details in next section.

### Full API Example

For the sake of clarity I'll try post an example of using every component with this new API.

```hbs
{{#let (file-queue fileAdded=this.addFile fileRemoved=this.removeFile) as |queue|}}
  {{#each queue.files as |file|}}
    <span data-test-file>
      {{file.name}}
    </span>
    <button type='button' data-test-remove {{on 'click' (fn queue.remove file)}}>
      x
    </button>
  {{/each}}

  <label>
    <input
      type='file'
      {{queue.selectFile filter=this.filter filesSelected=this.selectFiles}}
      hidden
    />
    Select File
  </label>

  <FileUpload
    @queue={{queue}}
    @filter={{this.filter}}
    @filesSelected={{this.selectFile}}
    {{!-- deprecated --}}
    @onSelect={{this.filterSelectedFiles}}
    @onFileAdd={{this.performUpload}}
  >
    <a>choose files</a>
  </FileUpload>

  <FileDropzone
    {{!-- we probably need to add --}}
    @queue={{queue}}
    @filter={{this.filter}}
    @filesSelected={{this.selectFile}}
    {{!-- deprecated --}}
    @onDrop={{this.filesDropped}}  {{!-- and many more drag and drop callbacks --}}
    @onFileAdd={{this.performUpload}}
  />
{{/let}}
```

I think we should make the naming consistent and replace with deprecation all `onEvent` callbacks with `event`.

### Performing an upload

The currently documented way to integrate is for the consumer to pass their own callback to `onFileAdd` on one of the components and then call `file.upload()` within it.

This much nicer API means we can recommend users instead attach the same callback to `fileAdded` on the `file-queue` helper.

### Final thoughts

How can I help with this? 

- [ ] perform at least the minimum update to `<FileDropzone>`
- [ ] make code/test/documentation contributions. Much of the docs needs updating to support this
- [ ] (not ready yet but) write an Upgrading to 5.0 doc

gossi:
Thanks @gilest for your feedback. Let me address some of those.

> Appreciate that TS is very nice and helpful – and I don't want to discourage you – but it's a bit presumptuous to include without at least discussing first. Would also prefer to make this change in a single PR rather than bundle it with an API update.

Yeah, I tried to use JS in the first place. But was too often distracted by fields that were "magically" inserted and I needed clarity to access them. So I pulled in TS. I saw discussion about using TS for this addon in the issues. I also thought about doing that in isolation as a "foundation" PR at first. But this PR would be to only install ember-cli-typescript and with subsequent PRs type one file after the other (or you lower the type-checks close to zero, when renaming existing `.js` to `.ts` files) - I've tried that as well and can tell now, this is not a good thing to do :/

> > With usage of TS I had to disable ember-cli-addon-docs (maybe it is time to move, ie. field-guide or docfy?)
> 
> Are you volunteering to take care of this? I don't want to release without a docsite and the current one hosts versioned docs back to 2.4.4 which we'd have to archive somehow.

Yes. It's done when it's documented. I looked into other options than ec-addon-docs, namely field-guide and docfy. Given the later plays well with TS (and probably auto-generates the API for components ?), this seems to be the go to option.

> > Use @ember/destroyables (depending on support matrix, this requires to ship the polyfill)
> 
> Yes we'll need `ember-destroyable-polyfill` as we currently support `v3.16` and above.

Ok, I had this in. Couldn't find the minimum supported ember-version and because the `ember-source` dependency was latest version, I removed it 🤦. Haven't checked `ember-try` config. Gonna bring it back 👍 

> 
> > Had to drop @onselect (couldn't rewire to the new API - I don't see this as an issue, the API was never public)
> 
> It was recently added and documented in #578. It has been published in betas but not an official release.
> 
> Looks like with this api `filter` takes care of filtering and `filesSelected` is available as a callback for file selection. Seems good 👍🏻 Someone will need to update the docs.
> 

Ah, I wasn't precise. It was never public as stable only tagged as beta, which according to semver is "here is a playground for a new API, we need feedback - for the time please don't rely on this". So, it wasn't released with _final_ API guarantees yet - that's my point.

> > I think, a lot of things can be marked deprecated for v5 and removed in v6
> 
> True. I would be happy to more aggressively update `<FileUpload>`. At least by removing the attribute arguments and using `...attributes`. This would be a very easy upgrade to document for users.
> 

I'm sorry to disappoint you but `...attributes` is already in use for this component and you cannot have two of them. That's why:

> I guess regardless I would recommend addon consumers use the `queue.selectFile` method over `<FileUpload>` anyway... The main outstanding issue we hear about is with DOM structure and inaccessibility of hidden file input.
> 


> > No API for the upload process, ie. (file-queue uploadStarted=... upload=...) or Queue.upload() - where does upload=() need to be attached? To (file-queue) or Queue?
> 
> Adding upload callbacks to the `file-queue` helper makes sense to me. This would need to be registered on the `Queue` instance like the other callbacks.
> 

Perhaps not needed at all. Since `upload()` is happening in userland and `filesSelected()` is notifying. This solved the problem I was having. Fine to move that into the waiting hall, less worries 😄 

> > I haven't touch drag'n'drop at all. I event kept Queue._addFiles() - which is sooo ready to go 😅
> 
> This can mostly be handled internally since the API changes are happening at the queue level, yeah?
> 
> We will need to at least allow `@queue` to be passed in as an arg though. See details in next section.

Yeah, I guess so. `@queue` would be the newest addition, while `@name` should receive a deprecation warning.

> ### Full API Example
> 
> For the sake of clarity I'll try post an example of using every component with this new API.
> 
> ```
> {{#let (file-queue fileAdded=this.addFile fileRemoved=this.removeFile) as |queue|}}
>   {{#each queue.files as |file|}}
>     <span data-test-file>
>       {{file.name}}
>     </span>
>     <button type='button' data-test-remove {{on 'click' (fn queue.remove file)}}>
>       x
>     </button>
>   {{/each}}
> 
>   <label>
>     <input
>       type='file'
>       {{queue.selectFile filter=this.filter filesSelected=this.selectFiles}}
>       hidden
>     />
>     Select File
>   </label>
> 
>   <FileUpload
>     @queue={{queue}}
>     @filter={{this.filter}}
>     @filesSelected={{this.selectFile}}
>     {{!-- deprecated --}}
>     @onSelect={{this.filterSelectedFiles}}
>     @onFileAdd={{this.performUpload}}
>   >
>     <a>choose files</a>
>   </FileUpload>
> 
>   <FileDropzone
>     {{!-- we probably need to add --}}
>     @queue={{queue}}
>     @filter={{this.filter}}
>     @filesSelected={{this.selectFile}}
>     {{!-- deprecated --}}
>     @onDrop={{this.filesDropped}}  {{!-- and many more drag and drop callbacks --}}
>     @onFileAdd={{this.performUpload}}
>   />
> {{/let}}
> ```
> 
> I think we should make the naming consistent and replace with deprecation all `onEvent` callbacks with `event`.

👍 

> ### Performing an upload
> 
> The currently documented way to integrate is for the consumer to pass their own callback to `onFileAdd` on one of the components and then call `file.upload()` within it.
> 
> This much nicer API means we can recommend users instead attach the same callback to `fileAdded` on the `file-queue` helper.

Yes, as said above, this needs a great documentation update.

> ### Final thoughts
> 
> How can I help with this?

What I can think of:

- Take care of drag'n'drop in a follow-up PR
- See what parts of the API can be deprecated; an issue as start and a follow-up PR
- Documentation as follow-up PR (I will help with that)
- Feel free to ping me on discord 👍 

What I will do:

- Address your PR comments
- At my work integration I've seen the queue files not being updated. That's when I went into vacation, need to figure out if this is happening in addon code or userland code
- Documentation as follow-up PR

But much likely this will all begin to happen next year 🎉

addon/components/file-upload.hbs
```diff
@@ -2,12 +2,7 @@
   for={{this.for}}
   class="file-upload"
   ...attributes
-  {{update-queue
-    this.queue
-    multiple=@multiple
-    disabled=@disabled
-    onFileAdd=@onFileAdd
-  }}
+  {{did-insert this.attachListeners}}
```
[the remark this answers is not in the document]
gossi: makes sense, fully agree to keep dependencies to a minimum. Didn't thought about that usage 👍

addon/queue.ts
```diff
@@ -0,0 +1,294 @@
...
+  }) {
+    this.name = name;
+    this.fileQueue = fileQueue;
+  }
+
+  // @TODO this is called from nowhere ?!?
+  // destroy() {
+  //   this.distinctFiles.forEach((file) => (file.queue = undefined));
+  //   this.distinctFiles.clear();
+  // }
```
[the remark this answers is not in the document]
gossi: Yeah, might be a "thing from the past". Also I share your understanding. We do have destroyables now, so gonna use them instead 👍

addon/services/file-queue.ts
gossi: Deprecate, because they are part of public API and potentially are used as part of this "super-global" queue. They should receive deprecation warnings 👍

addon/components/file-upload.ts
gossi: Yeah, deprecation warnings as they are still part of the public API. A contract for the addon to stick to until v6 then.

addon/file.ts
```diff
@@ -0,0 +1,323 @@
...
+
+  /** A unique id generated for this file. */
+  get id(): string {
+    return this.#id;
+  }
+
+  /** The file name */
+  get name(): string {
+    // @TODO: this "default" name here looks wrong - the tests wanted it ?!?
+    return this.#blob?.name ?? 'blob';
```
gossi, line 107: I added this default name `blob` here. It seemed necessary because tests demanded it, but I have no idea where else it was coming from?

I'd rather have no `blob` default name and change the tests? Dunno if it'll break some consumers app.

addon/file.ts
```diff
@@ -0,0 +1,323 @@
...
+    return upload(this, url, options, (request, opts) => {
+      // Build the form
+      const form = new FormData();
+
+      for (const key of Object.keys(opts.data)) {
+        if (key === opts.fileKey) {
+          // @TODO: Am I doing this right here?
+          form.append(key, this.#blob, this.name);
+        } else {
+          form.append(key, opts.data[key]);
```
gossi, line 240: Need help on the first `form.append()`. They were both thhe same code in current master but I was changing the one in the first if condition until tests were happy. Am I doing this right?

addon/queue.ts
```diff
@@ -0,0 +1,286 @@
...
+   * If the queue belongs to a top level collection,
+   * photos, the good name for this queue may be `"photos"`.
+   *
+   * If you're uploading images to an artwork, the
+   * best name would incoporate both `"artworks"` and
+   * the identifier of the artwork. A good name for this
+   * queue may be `"artworks/{{id}}/photos"`, where `{{id}}`
+   * is a dynamic segment that is generated from the artwork id.
+   */
+  name: QueueName;
```
gossi, line 52: should this be readonly?

gossi:
Ok, fixed the bug I've been seeing, addressed your requested changes and made some cleanup.

- I put `@deprecated` everywhere (at least in docblocks), so in a follow-up PR this can be done properly with `deprecate()` imported from ember and proper error messages.
- I left some questions above in my code where I needed to make little changes to make the tests go green. However this was kinda a blind change.

These question plus how to deal with `@onSelect`?  I'm in favor of removing it (as I kinda killed its purpose but brought in a substitue).

That would be the last thing. Apart from that, if there is nothing left from your side, this PR is already good looking and ready to go?

gossi:
@gilest Any updates on this? Anything for me to do? Would like to see this merged and published :)

gossi:
rebased and fixed the linting... which needs approval again (can't this run automatically?)

tests/helpers/file-queue-helper-test.js
```diff
@@ -0,0 +1,162 @@
...
+    assert.dom('[data-test-file]').hasText('dingus.txt');
+
+    await settled();
+    assert.dom('[data-test-file]').doesNotExist();
+
+    // second upload (retention teset)
+    // to check `Queue.flush()` will keep auto-tracking intact
+    selectFiles('input[type=file]', new File([], 'dingus.txt'));
+    await waitFor('[data-test-file]');
+    assert.dom('[data-test-file]').hasText('dingus.txt');
```
ijlee2, line 157: @gossi Would writing `await selectFiles(...)` and removing `await waitFor(...)` help with the failing test in CI?

addon/helpers/file-queue.ts
```diff
@@ -0,0 +1,65 @@
...
+export default class FileQueueHelper extends Helper implements QueueListener {
+  @service declare fileQueue: FileQueueService;
+
+  declare args: FileQueueArgs;
+
+  compute(_positional: unknown[], args: FileQueueArgs) {
+    this.args = args;
+    const queue = args.name
+      ? this.fileQueue.findOrCreate(args.name)
+      : this.fileQueue.findOrCreate(DEFAULT_QUEUE);
```
ijlee2, line 47: @gossi Here, I think you can refactor code.

```suggestion
    const queue = this.fileQueue.findOrCreate(args.name ?? DEFAULT_QUEUE);
```

gossi:
Thank you @ijlee2 for the comments.

@gilest I was looping Isaac in, because I have no idea why the two tests are failing on CI. The important piece here is, that they append `--launch chrome` argument to it, which actually does _something_ to it - yet we both do not really know why.

Any idea?

addon/file.ts
gilest: Right, I see the problem here.

I've renamed the internal `File` class to `UploadFile` – been meaning to do this because it conflicts with native js/browser `File` class.

This has allowed me to refactor the Blob/File handling throughout the internals. Please [see my commit here](https://github.com/adopted-ember-addons/ember-file-upload/commit/cab88eb5f62394d1aac98a25671586f78e8af18e) which you could cherry-pick if you like.

Essentially there had been some shortcuts taken and overuse of the `fromBlob` factory. I've refactored so that the `UploadFile` instance always has a native `File` which makes the internals much simpler. Hope this helps.

addon/file.ts
gilest: Hopefully resolved above ☝🏻 . Didn't require a code change.

addon/queue.ts
gilest: Yes it shouldn't modified.

tests/helpers/file-queue-helper-test.js
gilest: I see what you're trying to do here – but even locally for me the timing of this test seems unreliable.

When a file is selected and immediately uploaded the `Queue.flush` method should run, but it all happens in a very short time.

Suspect glimmer is optimising for minimal renders and deciding not to update the DOM with every change here. I got it to pass more often by adjusting the timing of mirage but it still wasn't and acceptable failure rate.

Could you explain clearly what you're trying to test? You're checking rendered page state since you mention autotracking, yeah? Is there a simpler way to test this?

addon/file.ts
gossi: Aaaah, very nice addition. When I realized I'm in conflict with `File` then I aliased this one to `UploadFile`, too already. Very much appreciated change 👍 

And also answers my two questions, thank you :)

tests/helpers/file-queue-helper-test.js
gossi: Well, I wanna test the "uploading" state and that the UI is reacting to that state change.

When doing `await selectFiles()` this will be skipping over the `uploading` state to whatever comes after. So, not awaiting this and `waitFor()` the element that should pick up the UI change kinda seemed the way to go - and for what it's worth, it worked for the situations I used that. I also had tests, that weren't possible this way.

I also used a waiting timer in the BE (I used polly with a delay in our app code), that didn't worked at all, the `waitFor()` was much more reliable than the timer. My very assumptive exaplanation is that the runloop is queuing those steps in and allows testing in that order, if they are very laboratory and an execution in this order can be guaranteed.

That said, I'm happily taking advice for this one. Maybe we can provide better test adapters? (for mirage and polly?)

gossi: I've used a timeout for this case here. Let's see if CI is happy with that. Also this is not a problem introduced as part of this PR, more that this PR uncovered this issue.

As a matter of that, I created an issue to discuss ideas for solving that in the future #641 and keep the scope of this PR to "make-the-CI-green"

addon/queue.ts
```diff
@@ -0,0 +1,281 @@
...
+
+        let uploadFile;
+        if (file instanceof File) {
+          uploadFile = new UploadFile(file, FileSource.Browse);
+        }
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        else if (file instanceof Blob) {
+          uploadFile = UploadFile.fromBlob(file, FileSource.Browse);
+        }
```
gossi, line 260: @gilest I made this case-switch after your changes on `UploadFile` constructor. That was necessary to make `selectFiles()` from `tests-support` work.

Is this as you intended it? Wdyt?

gilest (approved):

PR #654, opened 2022-01-21
Bind FileUpload component to the queue with a modifier. Drop `@ember/render-modifiers`

gilest:
Previously, components that are no longer rendered would still be attached to the queue as a listener.

Adding the tests is how I verified that this issue exists. Included tests should ensure this behaviour is maintained.

Allows us to drop the `@ember/render-modifiers` package also.

### DIFF
--- a/addon/components/file-upload.ts
+++ b/addon/components/file-upload.ts
@@ -1,7 +1,6 @@
 import Component from '@glimmer/component';
 import { inject as service } from '@ember/service';
-import { action } from '@ember/object';
-
+import { modifier } from 'ember-modifier';
 import UploadFile from 'ember-file-upload/upload-file';
 import Queue from '../queue';
 import FileQueueService, { DEFAULT_QUEUE } from '../services/file-queue';
@@ -105,21 +104,17 @@ export default class FileUploadComponent extends Component<FileUploadArgs> {
       return this.args.queue;
     }
 
-    if (this.args.name) {
-      return this.fileQueue.findOrCreate(this.args.name);
-    }
-
-    return this.fileQueue.findOrCreate(DEFAULT_QUEUE);
+    return this.fileQueue.findOrCreate(this.args.name ?? DEFAULT_QUEUE);
   }
 
   get for() {
     return this.args.for ?? `file-input-${guidFor(this)}`;
   }
 
-  @action
-  attachListeners() {
+  bindListeners = modifier(() => {
     this.queue.addListener(this);
-  }
+    return () => this.queue.removeListener(this);
+  });
 
   fileAdded(file: UploadFile) {
     this.args.onFileAdd?.(file);
--- a/tests/integration/queue-listeners-test.js
+++ b/tests/integration/queue-listeners-test.js
@@ -1,0 +1,89 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render, settled } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+import { selectFiles } from 'ember-file-upload/test-support';
+import { tracked } from '@glimmer/tracking';
+
+module('Integration | queue listeners', function (hooks) {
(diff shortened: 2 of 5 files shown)


addon/components/file-upload.ts
```diff
@@ -105,21 +104,17 @@ export default class FileUploadComponent extends Component<FileUploadArgs> {
...
     return this.args.for || `file-input-${uuid.short()}`;
   }
 
-  @action
-  attachListeners() {
+  bindListeners = modifier(() => {
     this.queue.addListener(this);
-  }
+    return () => this.queue.removeListener(this);
+  });
```
gilest, line 117: Is there any typing information I should include here? I don't think the modifier needs to accept arguments since the whole object (this) is what's bound/unbound.

gilest:
@gossi Would appreciate your review

addon/components/file-upload.ts
gossi: As there are no args involved... enjoy the clean look :)

tests/integration/queue-listeners-test.js
```diff
@@ -0,0 +1,89 @@
...
+
+    assert.verifySteps([
+      'first component: first.txt',
+      'second component: second.txt',
+    ]);
+  };
+
+  test('attaches and detaches filesSelected from queue', async function (assert) {
+    assert.expect(3);
+    setupState.bind(this)();
```
gossi, line 36: clever :)

gossi (approved):
Dang, I missed that one. Well spotted, elegant solution :)

tests/integration/queue-listeners-test.js
```diff
@@ -0,0 +1,89 @@
...
+    class State {
+      @tracked firstComponent = true;
+    }
+
+    this.state = new State();
+    this.queue = fileQueueService.create('listener-test');
+  };
+
+  const assertOnlyRenderedComponentCallbacks = async function (assert) {
+    await selectFiles('input[type=file', new File([], 'first.txt'));
```
gossi, line 23:
```suggestion
    await selectFiles('input[type=file]', new File([], 'first.txt'));
```

tests/integration/queue-listeners-test.js
```diff
@@ -0,0 +1,89 @@
...
+    this.state.firstComponent = false;
+    await settled();
+    await selectFiles('input[type=file', new File([], 'second.txt'));
```
gossi, line 26:
```suggestion
    await selectFiles('input[type=file]', new File([], 'second.txt'));
```

my inner monk wants the square bracket to be closed 😂

gilest: Good catch 🙇🏻

gilest:
I can't get the Embroider tests to pass unless `@ember/render-modifiers` is included at least as a `devDependency`.

I don't understand why as that addon is quite simple and I don't think we're using anything that it exports.

Have tried adding/pinning a number of related dependencies but so far no luck.

This is a pain as this is kind of blocking the API upgrade of FileDropzone.

jelhan:
> I can't get the Embroider tests to pass unless `@ember/render-modifiers` is included at least as a `devDependency`.
>
> [...]
>
> This is a pain as this is kind of blocking the API upgrade of FileDropzone.

What is the drawback of doing so? I get that it is not optimal. But it should not leak to public API as long as it is in `devDependency` only. I think it is okay to leave it as `devDependency` for now to not block the next release.

As it is okay to have it as a `devDependency` only, I'm wondering if it is caused by an addon, we consume in testing only. Or maybe an issue with Ember itself. :thinking: If you want to invest more time into this one, I would recommend to try to come up with a reproduction. Creating a new addon and adding dependencies as used here step by step until it breaks. But that could be very time consuming. Especially if the issue is only triggered by specific versions of dependencies or depends on how they are used.

gilest:
Never mind @jelhan. Solved it by pinning `ember-auto-import` to v2. (Saw this done in #675)

PR #655, opened 2022-01-21
Use guidFor in place of uuid

gilest:
Believe this is a safe replacement. Don't see why we need to ship a custom uuid function.

[Docs for guidFor](https://api.emberjs.com/ember/release/functions/@ember%2Fobject%2Finternals/guidFor).


gilest:
@gossi would appreciate your review

gossi (approved):
sounds good 👍

Issue #641, opened 2022-01-17
Improve test-support


gossi:
When working on #620 my new discovered some shortcomings of the current test-support. This is a spin-off of the mentioned PR to discuss ideas to improve it :)

## Utilize Test-Support ;)

Currently, the test-support for mirage is part of the public API for production code. It should be moved `test-support` to not pollute prod builds with that chunk of code.

Specifically: move `addon/mirage` to somewhere into `addon-test-support`

## Multiple Adapters

At the moment, `ember-file-upload` ships with a handler for mirage. Would be nice to have support for polly.js, too.

Even better, I think equally imporant are the test-support utilities, such as `extractFormData()` and `extractFileMetadata()`. They are very much suitable for integration tests, ie.

```
// understand this as pseudo-code, I'm not writing this from memory (API and names are likely to be different)
test('it uploads (with polly)', async function(assert) {
  this.server.polly.post('/photos/new', (req, res) => {
    extractFileMetadata(req.responseBody);
  });
});
```

As I created those tests myself without/prior to knowing those utils exists, they are likely to work with either polly as well as mirage :)

## Imperatively Control the Upload Progress

Writing UIs around upload is definitely part of userland code. Giving developers a good way to test their UIs is key. To test what happens in each `file.state`? Using `await selectFiles()` would skip the `uploading` state and would require devs to either `waitFor()` of some dom node to appear or set some timeout on the request itself. Both are not reliable and have high potential to cause flaky tests, [read more in this discussion](https://github.com/adopted-ember-addons/ember-file-upload/pull/620#discussion_r783994808).

An idea would be to have an imperative API to programmatically advance the `state` and be _in control_. Basically, a deferred promise - this is _shim_ code:

```
test('it uploads (with polly)', async function(assert) {
  const uploadHandler = new PollyUploadHandler();
  this.server.polly.post('/photos/new', async (req, res) => {
    return uploadHandler.handle((req, res) => {
      // business logic code
    });
  });

  await render(hbs`...`);

  const upload = selectFilesForHandler('input[type="file"]', new File(...));

  uploadHandler.startUpload(upload); // moves `state` into `uploading`

  assert.dom('...').hasText('uploading...'); // asserts UI state for uploading

  uploadHandler.success((req, res) => { 
    // possible business logic code
  });

  // --- OR ---

  uploadHandler.fail(); // example for not making use of business logic
  
  assert.dom('...').hasText('upload failed'); // assert UI for `error` state
});
```

[...]

PR #664, opened 2022-01-21
Clear mirage deprecations

jrjohnson:
In v3.x functions from miragejs will no longer be exported by mirage and
must be pulled directly from miragejs (which I've added as a development dependency here).

The configuration system has been modified to match the upstream
expectations and by more explicit as well.

It seems like `miragejs` would probably need to get pushed out as a dependency or peedDependency in order to be imported from the `addon` directory, however I'm not sure what impact that would have on the build of the consuming app, how to document peerDependencies (as they aren't used much), or if there is a way around this. 

I think the best path may be to remove `Response` and ship it as a local object which meets the interface needs of mirage, that's not usually a great idea though. Open to all suggestions!

### DIFF
--- a/addon/mirage/index.js
+++ b/addon/mirage/index.js
@@ -1,6 +1,11 @@
 import RSVP from 'rsvp';
-import Response from 'ember-cli-mirage/response';
+
 import { extractFormData, extractFileMetadata } from './utils';
+import {
+  macroCondition,
+  dependencySatisfies,
+  importSync,
+} from '@embroider/macros';
 
 const NETWORK = {
   wired: 50000, // 500 Mb/s
@@ -14,61 +19,73 @@ const NETWORK = {
 };
 
 export function upload(fn, options = { network: null, timeout: null }) {
-  return function (db, request) {
-    let speed = Infinity;
+  if (
+    macroCondition(
+      dependencySatisfies('miragejs', '*') &&
+        dependencySatisfies('ember-cli-mirage', '*')
+    )
+  ) {
+    const { Response } = importSync('miragejs');
+    return function (db, request) {
+      let speed = Infinity;
 
-    if (NETWORK[options.network]) {
-      speed = NETWORK[options.network] * 1024;
-    }
+      if (NETWORK[options.network]) {
+        speed = NETWORK[options.network] * 1024;
+      }
 
-    let { file, data } = extractFormData(request.requestBody);
-    let loaded = 0;
-    let total = file.value.size;
+      let { file, data } = extractFormData(request.requestBody);
+      let loaded = 0;
+      let total = file.value.size;
 
-    return new RSVP.Promise((resolve) => {
-      let start = new Date().getTime();
-      let metadata = extractFileMetadata(file.value);
+      return new RSVP.Promise((resolve) => {
+        let start = new Date().getTime();
+        let metadata = extractFileMetadata(file.value);
 
-      let upload = () => {
-        let timedOut =
-          options.timeout && new Date().getTime() - start > options.timeout;
-        if (timedOut || loaded >= total) {
-          request.upload.onprogress({
-            lengthComputable: true,
-            total,
-            loaded: Math.min(loaded, total),
-          });
-
-          metadata
-            .then((metadata) => {
-              request.requestBody = Object.assign(
-                {
-                  [file.key]: metadata,
-                },
-                data
-              );
-              if (timedOut) {
-                resolve(new Response(408));
-                return;
-              }
-
-              resolve(fn.call(this, db, request));
-            })
-            .catch((error) => {
-              resolve(new Response(500, {}, { error: error.message }));
+        let upload = () => {
+          let timedOut =
+            options.timeout && new Date().getTime() - start > options.timeout;
+          if (timedOut || loaded >= total) {
+            request.upload.onprogress({
+              lengthComputable: true,
+              total,
+              loaded: Math.min(loaded, total),
             });
-        } else {
-          request.upload.onprogress({
-            lengthComputable: true,
-            total,
-            loaded,
-          });
 
-          loaded += speed / 20;
-          setTimeout(upload, 50);
-        }
-      };
-      upload();
-    });
-  };
+            metadata
+              .then((metadata) => {
+                request.requestBody = Object.assign(
+                  {
+                    [file.key]: metadata,
+                  },
+                  data
+                );
+                if (timedOut) {
+                  resolve(new Response(408));
+                  return;
+                }
+
+                resolve(fn.call(this, db, request));
+              })
+              .catch((error) => {
+                resolve(new Response(500, {}, { error: error.message }));
+              });
+          } else {
+            request.upload.onprogress({
+              lengthComputable: true,
+              total,
+              loaded,
+            });
+
+            loaded += speed / 20;
+            setTimeout(upload, 50);
+          }
+        };
+        upload();
+      });
+    };
+  } else {
+    throw new Error(
+      'You must add ember-cli-mirage and miragejs to your app to use this helper.'
+    );
+  }
 }
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "@glimmer/component": "^1.0.4",
     "@glimmer/tracking": "^1.0.4",
     "broccoli-funnel": "^3.0.8",
+    "ember-auto-import": "^2.0.0",
     "ember-cli-babel": "^7.26.6",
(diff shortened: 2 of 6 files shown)


jelhan:
Thanks a lot for your contribution!

> It seems like `miragejs` would probably need to get pushed out as a dependency or peedDependency in order to be imported from the `addon` directory, however I'm not sure what impact that would have on the build of the consuming app, how to document peerDependencies (as they aren't used much), or if there is a way around this.
> 
> I think the best path may be to remove `Response` and ship it as a local object which meets the interface needs of mirage, that's not usually a great idea though. Open to all suggestions!

An alternative would be to pull the Mirage support out in a separate package before landing this changes as discussed in https://github.com/adopted-ember-addons/ember-file-upload/issues/641. This would make it less risky to include `miragejs` as a dependency because we could assume that a consumer of that package is using Mirage. Doing so we could avoid the complexity of an optional peerDependency.

Using [Embroider macros](https://github.com/embroider-build/embroider/tree/main/packages/macros#embroidermacros) `dependencySatisfies` and `importSync` might be an alternative. Doing so should allow us to only import `miragejs` if consuming application has a dependency on that package. This is included as one example use case in the official documentation for Embroider macros: https://github.com/embroider-build/embroider/tree/main/packages/macros#dependencysatisfies But in my experience Embroider macros add quite some complexity. There is a reason why it hasn't reached a stable `1.0.0` release yet.

jrjohnson:
Ok, I took a pass at doing this with macros, however due to https://github.com/embroider-build/embroider/issues/199 this will file unless both `ember-cli-mirage` and `mirages` are placed into `peerDependencies`. This code seems like a fairly ideal example of why that shouldn't be true, but I'm not sure how possible it is to update embroider to cover this situation. I wanted to push it as an example, but maybe pulling this into it's own package is the only way to move forward.

addon/mirage/index.js
```diff
@@ -14,61 +19,73 @@ const NETWORK = {
 };
 
 export function upload(fn, options = { network: null, timeout: null }) {
-  return function (db, request) {
-    let speed = Infinity;
+  if (
+    macroCondition(
+      dependencySatisfies('miragejs', '*') &&
```
SergeAstapov, line 24: @jrjohnson could you please add to the `package.json` following meta info:

```json
  "peerDependencies": {
     "ember-cli-mirage": "*",
     "miragejs": "*"
   },
   "peerDependenciesMeta": {
     "ember-cli-mirage": {
       "optional": true
     },
     "miragejs": {
       "optional": true
     }
   },
```

You may see an example in https://github.com/miragejs/ember-cli-mirage/pull/2297

SergeAstapov (approved):

addon/mirage/index.js
jrjohnson: Thanks! That seems to maybe have done the trick.

gilest:
> Not sure if they need a restart or more attention

Have restarted the CI for you 🤞🏻

SergeAstapov:
Looks like something is still wrong with the import:

```
Built project successfully. Stored in "/tmp/tests-dist-2022024-1673-4hkmrc.cqznq".
82
not ok 1 Chrome 97.0 - [159 ms] - Integration | Helper | file-queue: filter is triggered when selecting files
83
    ---
84
        actual: >
85
            null
86
        stack: >
87
            Error: Could not find module `miragejs` imported from `(require)`
88
                at missingModule (http://localhost:7357/assets/vendor.js:256:11)
89
                at findModule (http://localhost:7357/assets/vendor.js:267:7)
90
                at requireModule (http://localhost:7357/assets/vendor.js:33:15)
91
                at Object.../../../externals/miragejs.js (http://localhost:7357/assets/chunk.c4aae3125db89d847f99.js:4207:18)
92
                at __webpack_require__ (http://localhost:7357/assets/chunk.a91ecf9989421fc5d8e6.js:1746:41)
93
                at upload (http://localhost:7357/assets/chunk.c4aae3125db89d847f99.js:851:157)
94
                at Server.routes (http://localhost:7357/assets/chunk.c4aae3125db89d847f99.js:3134:96)
95
                at Server.loadConfig (http://localhost:7357/assets/chunk.c433ad60b4345e672592.js:31746:14)
96
                at Server.config (http://localhost:7357/assets/chunk.c433ad60b4345e672592.js:31662:14)
97
                at new Server (http://localhost:7357/assets/chunk.c433ad60b4345e672592.js:31449:10)
98
        message: >
99
            beforeEach failed on filter is triggered when selecting files: Could not find module `miragejs` imported from `(require)`
100
        negative: >
101
            false
102
        browser log: |
103
    ...
```

gilest:
I ran the embroider test locally to see what I can debug.

Quite interesting, `ember-cli-mirage` is included in `require.entries` but `miragejs` is not. [Full entries dump here](https://gist.github.com/gilest/a4301eba3bc366c6991fd48a2f61bb3b).

Presumably something to do with the way/format `miragejs` are configured – I'm not very familiar with these.

So (besides still using a deprecated re-export) it actually does work with:

```js
const { Response } = importSync('ember-cli-mirage');
```

We could even move forward with that since it's deprecated until v3 (although not ideal).

Thoughts @SergeAstapov @jrjohnson

SergeAstapov:
@gilest tried to debug on my local and can't see why embroider (or `ember-auto-import`) does not do the job properly.

Tried to use `ember-cli-mirage@3.0.0-alpha.1` with `@embroider/macros@1.0.0` but no luck

jelhan:
I think I have seen an issue, which sounds similar if sub-dependencies pull in older versions of Embroider. In that case it was fixed by pinning recent Embroider version using yarn resolutions feature. Maybe something you want to try out here as well.

jrjohnson:
I just finished creating a simple as possible reproduction which [also fails](https://github.com/jrjohnson/test-miragejs-import/runs/4962685303?check_suite_focus=true#step:6:19) with a brand new v4 addon and just started a [discord conversation](https://discord.com/channels/480462759797063690/568935504288940056/936158726526763008). I'm at a loss.

jrjohnson:
Well moving `ember-auto-import` to dependencies has fixed the embroider builds, however the normal build is now broken. Ugh.

SergeAstapov:
> Well moving `ember-auto-import` to dependencies has fixed the embroider builds

I think we stepped into grey zone of eai + `@embroider/macros` and not yet v2 addon.

I wonder if failure caused by any < v1.12 `ember-auto-import@1` somewhere in the tree

gilest:
This is painful ❤️  – what about moving it to dependencies *and* pinning it 😅

SergeAstapov:
> This is painful ❤️ – what about moving it to dependencies _and_ pinning it 😅

this would be major version bump as consuming apps would have to migrate to e-a-i@2.

Alternative option is to add e-a-i@2 to dependencies only in `embroider-safe` and `embroider-optimized` scenarios to make ci green 🙈

gilest:
> this would be major version bump as consuming apps would have to migrate to e-a-i@2.

All current changes are being bundled into v5 until we can stand up a new docsite anyway.

> Alternative option is to add e-a-i@2 to dependencies only in embroider-safe and embroider-optimized scenarios to make ci green 🙈

😅  wouldn't this cause problems for consumers using embroider?

SergeAstapov:
> 😅 wouldn't this cause problems for consumers using embroider?

Not sure 100% tbh, I would check in real app to be confident, however high chances you are right.

jrjohnson:
VICTORY 😤!!!  In all my messing around I'd left EAI at 2.0 instead of ^2.0. Now that it's allowed to drift to 2.4 everything works as intended. I'm so happy I could scream.

SergeAstapov:
> VICTORY 😤!!! In all my messing around I'd left EAI at 2.0 instead of ^2.0. Now that it's allowed to drift to 2.4 everything works as intended. I'm so happy I could scream.

oh, so outdated version of EAI was the reason? can we now move it back to `devDependencies` if so?

jrjohnson:
Unfortunately not @SergeAstapov, according to https://github.com/embroider-build/embroider/issues/1091 it has to be in `dependencies`.

SergeAstapov:
> Unfortunately not @SergeAstapov, according to [embroider-build/embroider#1091](https://github.com/embroider-build/embroider/issues/1091) it has to be in `dependencies`.

ahhhh, I see.
I thought when EAI listed in `devDependencies` it rolls in npm packages at build time but seem like there is limitation and it needs to be in `dependencies`.
Not totally clear why adding to `devDependencies` is not sufficient, but let it be that way

gilest:
Thank you both for your help with this <3

Issue #642, opened 2022-01-17
New Documentation


gossi:
As I "deactivated" `ec-addon-docs` in #620 - I've looked into alternatives. Solutions available in ember-space are:

- [guidemaker](https://github.com/empress/guidemaker)/[field-guide](https://github.com/empress/field-guide)
- [docfy](https://docfy.dev/)

Solutions "outside" embers are plenty, such as [docusaurus](https://docusaurus.io/). From the mentioned solutions I've worked with docfy and docusaurus. For this case, I can definitely vote against a solution outside ember. From the remaining I can say `docfy` is now-that-I-did-that-once great to do, I have no hands-on-experience with empress suite. Both are fine working with markdown files. Carrying over existing documentation will work, examples need to be transformed to the choosen-style.

I personally have done one project in docfy and would do that again for this project. I've used it in combination with [picocss](https://picocss.com/) and had great results.

Here is the plan to implement this:

- [x] 1. PR: Setup monorepo [1]
- [x] 2. PR: Setup docfy
- [x] 3. PRs: Update docs 
  - [x] #721
  - [x] Add public API docs #740
- [x] 4. Deploy new docsite (without overwriting legacy site)

[1]: Best practices for docfy is to have this in a monorepo and use one dedicated project for the docs, not to couple it to the tests. Second good point is to use it to prep for [addons in v2 format](https://www.kaliber5.de/en/blog/v2-addon_en) (which look good in monorepo, too).

**Automatic Docs Generation**

I've looked into automatic documentation generators. So far I've found nothing useful, ie. using [api-extractor](https://api-extractor.com/) that would automatically "connect"/be integrated into doc tools isn't there yet. Also api-extractor is questionable to play with [`exports`](https://nodejs.org/api/packages.html#main-entry-point-export) field in `package.json` that addon v2 are using. TS team is more on the side of _one_ export per package. There is [glimmer-docgen-typescript](https://github.com/josemarluedke/glimmer-docgen-typescript), but would only document components and leave out a couple of important things for this addon.
Given the API surface of this plugin is luckily quite small, writing this manually with a practial and applicable examples in mind seems the best to do and likely the fastest as well (anyway time is spend in setting up an automation that will take more time than doing it manually).

Even then, as docfy can read markdown files and api-extractor/api-documentor are able to produce them is still able to look for combination in the future.

WDYT ?

gilest:
> For this case, I can definitely vote against a solution outside ember.

[...]

PR #675, opened 2022-01-25
Move into a monorepo

gossi:
First part for #642 - move into a monorepo, here wo go.

There are a couple of config files to be updated to work with that:

- `.github/workflows/ci.yml` - I did that (let's see)
- `renovate.json` - untouched, need to see if this works with a monorepo (without a change). Anyway [ember-resources](https://github.com/NullVoxPopuli/ember-resources/blob/main/.github/renovate.json5) has a good renovate config for ember projects

What else to look into?


gilest:
I don't have any experience with monorepo so can't review the methodology 🤷🏻 

Tried to get the CI to pass (in #677)

- Pinning a few deps
- Regenerating the lockfile

But no luck. I'm out of ideas there.

A final thought – we need to consider how the affects our release process. Do we need to install [release-it-yarn-workspaces](https://github.com/rwjblue/release-it-yarn-workspaces)?

gossi:
@gilest 
Ah, was kinda doing what you did. I upgraded the types dep and put `yarn.lock` on a diet 😅 

Ad release: cd into and you are good: `cd ember-file-upload && yarn release-it` (I did the same with `ember-cite` the past week(s)).

@jelhan ah cool. Didn't know. From what I read, there are no issues with the current config _for now_ but ways in the future to extend that

With that and CI going green, I consider this good to go. Can you please review this?

gilest (approved):
Looks fine to me but I'm not an expert.

Some minor update to the CONTRIBUTING.md and RELEASE.md files would be appreciated.

package.json
```diff
@@ -1,161 +1,9 @@
...
-  "name": "ember-file-upload",
-  "version": "5.0.0-beta.3",
-  "description": "An uploader component for Ember apps",
-  "keywords": [
-    "ember-addon",
-    "file-upload"
+  "private": "true",
+  "workspaces": [
+    "ember-file-upload"
   ],
```
jelhan, line 5: Did you considered having the workspaces in a folder called `packages/` instead of the repository root? That's a pattern at least I have seen often. It also allows to use a wildcard:

```suggestion
  "workspaces": [
    "packages/*"
  ],
```

That folder could contain later `ember-file-upload` itself, the docs app and libraries like `ember-file-upload-mirage-handler`.

When migrating to v2 addon format, we could add a `test-packages/` or `test-apps/` folder and move the dummy app to that one.

package.json
```diff
@@ -1,161 +1,9 @@
...
-    "prettier": "^2.3.2",
-    "qunit": "^2.16.0",
-    "qunit-dom": "^2.0.0",
-    "release-it": "^14.0.0",
-    "release-it-lerna-changelog": "^4.0.0",
-    "sass": "^1.32.10",
-    "typescript": "^4.5.3",
-    "webpack": "5"
-  },
   "resolutions": {
```
jelhan, line 6: Currently `resolutions` are defined in `package.json` for workspaces root and in `package.json` for `ember-file-upload` package. Personally I find this difficult to reason about. I would recommend to merge them at one place for now.

jelhan (approved):
I have two minor comments. But overall this looks good to me.

I think in the mid-term we may want to iterate on the setup. E.g. if we introduce additional libraries, which are released as separate packages, we may want to move the release setup to workspace root to release everything together. Or we want to introduce one or more test apps when migrating to v2 addon format. But I think all of that could be done later when we face the need.

package.json
gossi: Yeah, there are no established practices around that, how ember addons in v2 are organized on a repo-level. What I can see in the wild:

1. `packages/*` with [ember-cite](https://github.com/gossi/ember-cite) or [ember-stargate](https://github.com/kaliber5/ember-stargate/tree/master/packages) (they could be at root level)
2. `$packageName` and `testing/*` with [ember-resources](https://github.com/NullVoxPopuli/ember-resources)
3. Or not a monorepo as mentioned here: https://www.kaliber5.de/de/blog/v2-addon_en

I'd personally leaning more towards the second option (maybe because I tried the first one?). Now this is becoming also more interesting with (one) more package(s) in this repo, but tests living maybe in the same testing-package? This is a new challenge with this project. In the end, doesn't matter too much I'd say. As that can be organized back and forth to find the best fit 👍

package.json
gossi: Oh yes you are right. I didn't pay attention. Great catch, will move those to workspace `package.json`.

gilest:
@gossi Do you have a chance to make these final changes and rebase this? Thanks 🙇🏻

gilest:
Thank you @gossi 🙇🏻

PR #676, opened 2022-01-25
Avoid mutating upload options

### LINKED ISSUE #674: Upload options are mutated
We have an upload component in one of our projects that accepts additional upload data.
When passing along additional upload data using the `{{hash}}` helper the `setting-on-hash` deprecation is trigged because `ember-file-upload` mutates `options.data` [here](https://github.com/adopted-ember-addons/ember-file-upload/blob/master/addon/system/upload.js#L40) (also in other places).
We could just pass in a shallow copy ourselves, but to me, it seems preferable that `ember-file-upload` does not mutate the passed in options object in the first place.

bertdeblock:
Closes #674.

### DIFF
--- a/addon/system/upload.js
+++ b/addon/system/upload.js
@@ -5,13 +5,17 @@ import { buildWaiter } from '@ember/test-waiters';
 
 const uploadWaiter = buildWaiter('ember-file-upload:upload');
 
+function clone(object) {
+  return object ? { ...object } : {};
+}
+
 function normalizeOptions(file, url, options) {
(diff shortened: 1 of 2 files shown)


jelhan (approved):

gilest (approved):

gilest:
Thanks for reporting and taking care of this @bertdeblock 

Released as part of `v5.0.0-beta.3`

bertdeblock:
Sweet, thanks for the quick follow-up @jelhan and @gilest!

PR #678, opened 2022-01-28
Update FileDropzone API

gilest:
Adds new arguments `queue` and `filter` to FileDropzone.

Includes basic test coverage of these new APIs while deprecated API tests still pass.

Adds typing to `FileDropzone` component and `DataTransferWrapper` internal utility. I could have taken this further but I think it's good to do this process incrementally. Honestly there's a lot to untangle in some of these internal utilities (regarding typing support).

Registers/unregisters the FileDropzone as an event listener on the queue in the method from #654.

### A note about naming

Following the work of @gossi in #620 had planned to name the callback args `dragEnter`, `dragLeave` and `drop` but there's an `ember-template-lint` rule specifically discouraging this.

Screen Shot 2022-01-29 at 12 21 55 PM

I don't like the names I have chosen – will welcome other suggestions of how we name the callback arguments but I would very much like it to be consistent across the whole addon.

I've decided to revert to using existing event names with the `on` prefix.


addon/components/file-dropzone.ts
```diff
@@ -0,0 +1,354 @@
...
+    //       file.name = filename;
+    //     }
+    //   };
+    //   /* eslint-disable no-console */
+    //   image.onerror = function (e) {
+    //     console.log(e);
+    //   };
+    //   /* eslint-enable no-console */
+    //   image.src = url;
+    // }
```
gilest, line 326: This code concerns me as I have found there is no test coverage of it and I'm not sure how to test it manually.

I'm worried we may have broken this functionality without realising already. However I did try to drag an image from another browser window in `v4.0.3` and it didn't work so 🤷🏻‍♂️ .

Options:
- Explicitly remove this as it was undocumented and untested. Can always be added back as an on-breaking enhancement
- Try and figure out how it works and support it now

addon/queue.ts
```diff
@@ -173,41 +169,6 @@ export default class Queue {
...
-
-          if (onFileAdd) {
-            next(onFileAdd, uploadFile);
-          }
-        }
-      }
-    }
-
-    return files;
-  }
```
gilest, line 209: :wave

tests/helpers/file-queue-helper-test.js
```diff
@@ -119,7 +119,7 @@ module('Integration | Helper | file-queue', function (hooks) {
       uploadHandler(function (/*schema, request*/) {
         // do sth
       }),
-      { timing: 2000 }
+      { timing: 200 }
```
gilest, line 122: Two seconds was longer than the whole suite on my local. I tried reducing it and it passes locally and in CI. Seems fine 🙃

addon/components/file-dropzone.ts
```diff
@@ -0,0 +1,354 @@
...
+
+    if (this.dataTransferWrapper) {
+      // @TODO stop filtering files based on the output of onDrop
+      // in  favor of `filter()` - it never was officially public API
+      const files =
+        this.args.onDrop?.(this.files, this.dataTransferWrapper) ?? this.files;
```
gilest, line 332: I left this in for this PR but am very keen to remove it along with the equivalent functionality from `FileUpload` in a subsequent PR.

jelhan:
I will try to do a full review later today. But thought that leaving some thoughts regarding naming early may make sense.

> A note about naming
> 
> Following the work of @gossi in #620 had planned to name the callback args dragEnter, dragLeave and drop but there's an ember-template-lint rule specifically discouraging this.

Did you considered to keep the `on*` naming convention? For example `onDragEnter` and `onDrop`.

Prefixing event handlers with `on` helps with readability in my opinion. It communicates to the reader that this is an event handler, which executes a passed in function when a specific situation happens. It is also seems to be kind of a standard among favorite Ember addons.

It also helps to discover the available event handlers - especially when looking at autocompletion in templates. I hope that we will get to that in the foreseeable future. First-class component templates and Glimmer component signature type RFCs are very promising in that regard.

addon/components/file-dropzone.ts
```diff
@@ -0,0 +1,354 @@
...
+   */
+  cursor?: 'link' | 'none' | 'copy' | 'move';
+
+  // actions
+  filter?: (file: UploadFile) => boolean;
+
+  /**
+   * Called when files have entered the dropzone.
+   */
+  filesEnter?: (
```
gossi, line 39: That name seems inconsistent to me with the rest, it may be:

- `enterFile(s)` - apply business logic to mutate the data when entering the zone
- `filesEntered` - only a notifier, that some files have entered the zone (but no chance of altering them)

addon/components/file-dropzone.ts
```diff
@@ -0,0 +1,354 @@
...
+    files: File[] | DataTransferItem[],
+    dataTransfer: DataTransferWrapper
+  ) => void;
+
+  /**
+   * Called when files have left the dropzone.
+   */
+  filesLeave?: (
```
gossi, line 47: same here as above, either `filesLeft` or `leaveFiles` ?

addon/components/file-dropzone.ts
```diff
@@ -0,0 +1,354 @@
...
+    files: File[] | DataTransferItem[],
+    dataTransfer: DataTransferWrapper
+  ) => void;
+
+  /**
+   * Called when file have been dropped on the dropzone.
+   */
+  filesDropped?: (files: UploadFile[]) => void;
```
gossi, line 55: That one is clear though 👍

addon/components/file-dropzone.ts
```diff
@@ -0,0 +1,354 @@
...
+
+interface FileDropzoneArgs {
+  queue?: Queue;
+
+  /**
+   * Whether users can upload content from websites by dragging images from
+   * another webpage and dropping it into your app. The default is `false`
+   * to prevent cross-site scripting issues.
+   * */
+  allowUploadsFromWebsites?: boolean;
```
gossi, line 22: Is that argument already present? It starts with a verb, so it indicates it is callable, ie `allowUploadsFromWebsites()` - it is a _question_ rather than a _fact_. As _fact_ I'd call it `uploadsFromWebsitesAllowed`.

tests/integration/components/file-dropzone-test.js
```diff
@@ -7,135 +7,223 @@ import {
...
-        <div class="supported">{{dropzone.supported}}</div>
-        <div class="active">{{dropzone.active}}</div>
-        <div class="queue-name">{{queue.name}}</div>
-      </FileDropzone>
-    `);
-
-    assert.dom('.supported').hasText('true');
-    assert.dom('.active').hasText('false');
-    assert.dom('.queue-name').hasText('test');
+  module('deprecated api', function () {
```
gossi, line 78: oooooh, that's a nice one to move that into a `module` and to have the chance to delete it at once later on 👍

gossi (commented):
I love to see all this classic code go and see it replaced with new code.

I didn't look into the details of drag'n'drop code (as I don't know the details) but more on a "highlevel" view. Left some comments about the naming.

Apart from that, I think, it is looking really good.

addon/components/file-dropzone.ts
gilest: This is an existing argument. I'd prefer not to change it for now but can revisit in future.

addon/components/file-dropzone.ts
gilest: I've now gone ahead and done this as part of this PR.

gilest:
@gossi @jelhan Thanks for your comments. I've opted to use the original callback names and the `on` prefix.

I think it's important that the callback names are consistent so I've created #701 to track.

Believe this is ready to merge and will do so in a day or two if I don't get further reviews.

PR #682, opened 2022-02-02
`UploadFile.size `always returns 0

jrjohnson:
Since #size defaults to 0 this getter was always returning zero and
never the size of the actual file. I'm not sure under what conditions
this value needs to be settable, but I've followed the pattern for
testing the name and fixed the getter.

Maybe this would be better done by setting the default value to null, but I'm not familiar enough with typescript to see how to get that done correctly.

### DIFF
--- a/addon/upload-file.ts
+++ b/addon/upload-file.ts
@@ -112,7 +112,7 @@ export default class UploadFile {
 
   /** The size of the file in bytes. */
   get size() {
-    return this.#size ?? this.file.size;
+    return this.#size || this.file.size;
   }
 
   set size(value) {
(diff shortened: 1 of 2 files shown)


gilest:
Thank you @bertdeblock

It's a shame this wasn't covered by a test already as it's a piece of core functionality.

> I'm not sure under what conditions this value needs to be settable

It's used internally when files are uploaded:
https://github.com/adopted-ember-addons/ember-file-upload/blob/4714e79dc153b9d7cb0fcdb551893aba398cf216/addon/system/upload.js#L83

> Maybe this would be better done by setting the default value to null

That's true but it also may have some unintended effects as the sizes are summed together etc. Feel like that may end up with users seeing `null` or `NaN` in their templates.

bertdeblock:
I think you meant to ping @jrjohnson 😬.

gilest:
> I think you meant to ping @jrjohnson 😬.

Apologies 😅  it's late here

jrjohnson:
Makes sense @gilest, thanks for the quick merge and release!

PR #684, opened 2022-02-03
Exclude miragejs from ember-auto-import

### LINKED ISSUE #683: Failed to import `miragejs`
`5.0.0-beta.5` breaks our build with the following error:

```
Build Error (WebpackBundler)

ember-file-upload tried to import "miragejs" in "ember-file-upload/mirage/index.js" but the package was not resolvable from /Users/bertdeblock/Code/bagaar/projects/mobility-buddy/mobility-buddy-frontend/node_modules/ember-file-upload


Stack Trace and Error Report: /var/folders/q5/sg7x51yn0v5d4snmcrmcxxk00000gn/T/error.dump.ea295ef192933e99591559f0aa01612f.log
```

Installing `miragejs` ourselves fixes the issue, but our project doesn't use / need MirageJS.

gilest:
Fixes #683

I created a local reproduction, checked the `ember-auto-import` docs and some [real world examples](https://github.com/embroider-build/embroider/tree/main/packages/macros#real-world-examples) from the `@embroider/macros` docs.

It seems like if we're conditionally importing modules using `importSync/macros` then we need to explicitly exclude them from `ember-auto-import`.

Have tested this manually with a minimal repro. This could be replicated by adding more `ember-try` scenarios and using `@embroider/macros` to gate some tests.

Plain app:
with-embroider

With mirage:
with-mirage

With embroider:
with-embroider

With embroider and mirage:
with-embroider-mirage

I'm clueless about the CI builds that are failing. It seems to be related to generating sourcemaps for `qunit-dom`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,6 +3,11 @@ const funnel = require('broccoli-funnel');
 
 module.exports = {
   name: require('./package').name,
+  options: {
+    autoImport: {
+      exclude: ['miragejs'],
+    },
+  },
   included(app) {
     let config = app.project.config();
 
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,6 +2,13 @@
 # yarn lockfile v1
 
 
+"@ampproject/remapping@^2.0.0":
+  version "2.0.4"
+  resolved "https://registry.yarnpkg.com/@ampproject/remapping/-/remapping-2.0.4.tgz#ab4b6d858526ebee0d6c3aa5c3fb56a941c0d7be"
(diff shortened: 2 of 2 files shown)


SergeAstapov (approved):

SergeAstapov:
@gilest thank you for digging into this! After having another thought, this makes sense to me.

MelSumner:
Any ideas why the embroider scenarios are failing? Do we have some deps to update?

gilest:
> Any ideas why the embroider scenarios are failing? Do we have some deps to update?

No idea. The deps are pretty fresh and I've already tried regenerating `yarn.lock`. Will try again just in case.

PR #706, opened 2022-02-22
Deprecate `FileUpload` component and `FileDropzone` attribute args

gilest:
Register all v5 deprecations until v6. This includes the entire `FileUpload` component.

### DIFF
--- a/ember-file-upload/addon/components/file-dropzone.ts
+++ b/ember-file-upload/addon/components/file-dropzone.ts
@@ -146,7 +146,7 @@ export default class FileDropzoneComponent extends Component<FileDropzoneArgs> {
         id: 'file-dropzone.accept',
         since: { enabled: 'v5.0.0' },
         until: 'v6.0.0',
-        url: 'https://github.com/adopted-ember-addons/ember-file-upload/blob/master/docs/validation.md',
+        url: 'https://ember-file-upload.pages.dev/docs/upgrade-guide#filedropzone-component',
       }
     );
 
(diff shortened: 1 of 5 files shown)


ember-file-upload/addon/components/file-dropzone.ts
```diff
@@ -129,6 +141,77 @@ export default class FileDropzoneComponent extends Component<FileDropzoneArgs> {
     window.document &&
     'draggable' in document.createElement('span'))();
 
+  constructor(owner: unknown, args: FileDropzoneArgs) {
```
gilest, line 144: Is checking args in the constructor the best way to do this?

ember-file-upload/addon/components/file-upload.ts
```diff
@@ -88,11 +91,26 @@ interface FileUploadArgs {
...
+    deprecate(
+      `\`<FileUpload>\` is deprecated. Use \`{{file-queue}}\` helper with \`{{queue.selectFile}}\` modifier.`,
+      false,
+      {
+        for: 'ember-file-upload',
+        id: 'file-upload',
+        since: { enabled: 'v5.0.0' },
+        until: 'v6.0.0',
+      }
+    );
```
gilest, line 111: There's a lot of deprecated arguments also – but I figured a single deprecation warning with a (future) link to good upgrade docs is the best way forward here.

ember-file-upload/addon/components/file-dropzone.ts
```diff
@@ -27,9 +31,20 @@ interface FileDropzoneArgs {
...
   cursor?: 'link' | 'none' | 'copy' | 'move';
 
+  /**
+   * Whether to add multiple files to the queue at once.
+   *
+   * If set to false only one file will be added when dropping mulitple files.
+   *
+   * @defaultValue true
+   */
+  multiple?: boolean;
```
gilest, line 46: Decided not to deprecate as I can't think of an easy way to do this in user implementation.

Issue #709, opened 2022-02-23
Regression with surrounding white-space


nwhittaker:
Given this snippet:

```hbs
Drag and drop a file or <FileUpload>select from your device</FileUpload>.
```

The white-space surrounding the component is rendered differently:

| Version | Rendered | Notes |
| - | - | - |
| 4.0.3 | "Drag and drop a file or select from your device." | Shows the expected behavior. |
| 4.1 and 5.0 betas | "Drag and drop a file or select from your device ." | A space is introduced before the trailing period. |

gilest:
Thank you for reporting this.

The FileUpload component is being deprecated in v5 because it's quite inflexible in terms of DOM/markup. This is a good example of why!

It has been replaced by the `queue.selectFile` modifier which can be applied to a file input. Documentation of this is WIP which is why 5.0 is still in beta.

See [example usage here](https://github.com/adopted-ember-addons/ember-file-upload/blob/master/ember-file-upload/tests/helpers/file-queue-helper-test.js#L84):

```hbs
{{#let (file-queue onFileAdded=this.addFile onFileRemoved=this.removeFile) as |queue|}}
  <label>
    <input type="file" {{queue.selectFile}} hidden>
    Select File
  </label>
{{/let}}
```

nwhittaker:
Thanks for the example. Unfortunately, my project's constraints are such that I was looking for a short-term fix where I could upgrade to v5 but still use the old `FileUpload` component.

PR #711, opened 2022-02-23
Remove whitespace when rendering FileUpload

nwhittaker:
Adds a test and proposes a possible fix for https://github.com/adopted-ember-addons/ember-file-upload/issues/709.

`{{~!~}}` syntax from https://discuss.emberjs.com/t/removing-whitespace-with-angle-bracket-syntax/17094/3.

### DIFF
--- a/ember-file-upload/addon/components/file-upload.hbs
+++ b/ember-file-upload/addon/components/file-upload.hbs
@@ -4,6 +4,7 @@
   ...attributes
   {{this.bindListeners}}
 >
+  {{~!~}}
   <input
     id={{this.for}}
     type="file"
@@ -15,6 +16,6 @@
     {{this.queue.selectFile filter=@filter onFilesSelected=@onFilesSelected}}
   />
 
-  {{yield this.queue}}
+  {{~yield this.queue~}}
 </label>
-
+{{~!~}}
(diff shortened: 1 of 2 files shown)


ember-file-upload/addon/components/file-upload.hbs
```diff
@@ -4,6 +4,7 @@
   ...attributes
   {{this.bindListeners}}
 >
+  {{~!~}}
```
gilest, line 7: Can you explain or reference what this is doing (beyond a forum post).

Not sure I'm comfortable adding this if it's not a documented feature of glimmer as it may break in future versions.

nwhittaker: My understanding is it's coming from handlebars, not Glimmer. It looks to be combining the syntax for a comment with the syntax for white-space control: https://handlebarsjs.com/guide/expressions.html#whitespace-control.

Alternatively you could remove the white-space in this template file manually, but I was afraid that would be easier to regress.

gilest: Ok fair enough. In the interest of easing migration I'll merge this. As I mentioned in the issue, though, this component is about to be deprecated anyway.

nwhittaker: Understood on the deprecation, but thanks for merging a stop-gap -- much appreciated!

Issue #712, opened 2022-02-24
Max file limit handling in v5


mkszepp:
I want to know, if there is planned a option for max file limit in `v5` (for example allow max 3 files to upload)

I know, that html out of the box doesn't support this yet, but there is possible to handle with listener `drop` (for drag'n/drop) and `onChange` on input field.

My plan was to use the callbacks `onDrop` / `onFilesSelected` to handle this, but after code check, i have seen, that this callbacks will be called after the queue insert, For this reason this callbacks are coming to late for me. This callbacks has also confused me a little bit :), because the callbacks are not after dropping / file selected, they are more `afterFilesAdded`. A callback like `onDrop` / `onFilesSelected` is for me before adding files in the queue or?

After this discovery i was going on to check if there is already a other callback, which i can use for my limitation and i have seen the `filter` callback, but in this callback i have the problem that there is missing the index + all files, and so there is also here not possbile to make a file limiation, before adding in queue.
I think, the solution would be, to add the index & all files as parameter in `filter`.
If this solution is ok, i can also help to extend this callbacks.

https://github.com/adopted-ember-addons/ember-file-upload/blob/8add316307645fc738acbcf502d4f05ca6e5df65/ember-file-upload/addon/components/file-dropzone.ts#L302-L322

https://github.com/adopted-ember-addons/ember-file-upload/blob/4a329e9b1c01e46a65b12b2b3691aae473cc58ce/ember-file-upload/addon/queue.ts#L208-L229

gilest:
Hey – thanks for reporting.

Would something like the following work?

```js
const MAX_FILES = 3;

export default class ExampleComponent extends Component {
  @service fileQueue;

  get queue() {
    return this.fileQueue.findOrCreate('your-queue-name');
  }

  @action
  enforceMaxFiles() {
    return this.queue.files.length < MAX_FILES;
  }
}
```

Obviously passing `this.enforceMaxFiles` as your `filter` callback.

Although this is limiting the files that can be uploading at any given time, rather than limiting the number of files that can be selected each drop/select.

> I think, the solution would be, to add the index & all files as parameter in filter.

This is a reasonable solution, anyway, and a non-breaking change.

mkszepp:
Hi @gilest,
thank for your answer.

Yes, this solution works, when i want add files until the limit is reached, but is there also a possiblity to block all selected images, when someone selects 4 images, but only 3 are allowed?
[...]

PR #731, opened 2022-03-10
Add `files` and `index` to `filter` callback

mkszepp:
Issue https://github.com/adopted-ember-addons/ember-file-upload/issues/712

### DIFF
--- a/ember-file-upload/addon/components/file-dropzone.ts
+++ b/ember-file-upload/addon/components/file-dropzone.ts
@@ -31,11 +31,7 @@ interface FileDropzoneArgs {
   cursor?: 'link' | 'none' | 'copy' | 'move';
 
   // actions
-  filter?: (
-    file: File,
-    files: File[],
-    index: number
-  ) => boolean;
+  filter?: (file: File, files: File[], index: number) => boolean;
 
   /**
    * Called when files have entered the dropzone.
--- a/ember-file-upload/addon/components/file-dropzone.ts
+++ b/ember-file-upload/addon/components/file-dropzone.ts
@@ -32,8 +32,8 @@ interface FileDropzoneArgs {
 
   // actions
   filter?: (
-    file: UploadFile,
-    files: File[] | DataTransferItem[],
+    file: File,
+    files: File[],
     index: number
   ) => boolean;
 
@@ -145,8 +145,8 @@ export default class FileDropzoneComponent extends Component<FileDropzoneArgs> {
     return this.args.multiple ?? true;
   }
 
-  get files(): File[] | DataTransferItem[] {
-    const files = this.dataTransferWrapper?.filesOrItems ?? [];
+  get files(): File[] {
+    const files = this.dataTransferWrapper?.files ?? [];
     if (this.multiple) return files;
 
     return files.slice(0, 1);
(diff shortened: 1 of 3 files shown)


mkszepp:
@gilest during my changes i have discoverd, that in the filter there was passed one time the file as type UploadFile & one time as file.

Maybe there would be better, to pass alway as File or what do you think?

https://github.com/adopted-ember-addons/ember-file-upload/blob/c91488022fe2580aa592cfa9203d5c9762b116cd/ember-file-upload/addon/queue.ts#L7-L12

https://github.com/adopted-ember-addons/ember-file-upload/blob/c91488022fe2580aa592cfa9203d5c9762b116cd/ember-file-upload/addon/components/file-dropzone.ts#L34

gilest:
> in the filter there was passed one time the file as type UploadFile & one time as file.
> 
> Maybe there would be better, to pass alway as File or what do you think?

Yes, I agree

ember-file-upload/addon/components/file-dropzone.ts
```diff
@@ -310,10 +314,13 @@ export default class FileDropzoneComponent extends Component<FileDropzoneArgs> {
 
   addFiles(files: File[] | DataTransferItem[]) {
     const addedFiles = [];
+    let index = 0;
     for (const file of files) {
       if (file instanceof File) {
         const uploadFile = new UploadFile(file, FileSource.DragAndDrop);
-        if (this.args.filter && !this.args.filter(uploadFile)) continue;
+        if (this.args.filter && !this.args.filter(uploadFile, files, index++)) {
```
gilest, line 321:
```suggestion
        if (this.args.filter && !this.args.filter(file, files, files.indexOf(file)) {
```

ember-file-upload/addon/queue.ts
```diff
@@ -6,7 +6,7 @@ import FileQueueService from './services/file-queue';
 
 export interface SelectFileModifierArgs extends ModifierArgs {
   named: {
-    filter?: (file: File) => boolean;
+    filter?: (file: File, files: FileList, index: number) => boolean;
```
gilest, line 9:
```suggestion
    filter?: (file: File, files: File[], index: number) => boolean;
```

ember-file-upload/addon/queue.ts
```diff
@@ -204,9 +204,10 @@ export default class Queue {
       }
 
       const selectedFiles: UploadFile[] = [];
+      let index = 0;
 
       for (const file of files) {
-        if (named.filter && !named.filter?.(file)) {
+        if (named.filter && !named.filter?.(file, files, index++)) {
```
gilest, line 210: `files` will be a FileList rather than `File[]`. Suggest something like this:

```js
      const { files: fileList } = event.target as HTMLInputElement;
      if (!fileList) {
        return;
      }

      const files = Array.from(fileList);
      const selectedFiles: UploadFile[] = [];

      for (const file of files) {
        if (named.filter && !named.filter?.(file, files, files.indexOf(file))) {
          continue;
        }
```

ember-file-upload/addon/components/file-dropzone.ts
```diff
@@ -31,7 +31,11 @@ interface FileDropzoneArgs {
   cursor?: 'link' | 'none' | 'copy' | 'move';
 
   // actions
-  filter?: (file: UploadFile) => boolean;
+  filter?: (
+    file: UploadFile,
+    files: File[] | DataTransferItem[],
```
gilest, line 36: I don't think we need to expose any DataTransfer items that aren't files here.

Suggestion:
```js
  // file-dropzone.ts
  get files(): File[] {
    const files = this.dataTransferWrapper?.files ?? [];
    if (this.multiple) return files;

    return files.slice(0, 1);
  }
```

gilest:
Thank you @mkszepp

PR #732, opened 2022-03-13
New Docs Site

gossi:
Heyo,

who ordered a new docs site? I have one for you :)

Basically, I took the website project from `ember-cite` copied it to `ember-file-upload`, adjusted the config a little and give it an intro screen, that's it.

Please have a look. Code wouldn't tell much, at best check it local or the screenshots below.

Home:
Bildschirmfoto 2022-03-13 um 17 49 16

Docs:
Bildschirmfoto 2022-03-13 um 17 48 39

PR #734, opened 2022-03-18
Generate new Docfy documentation website

gilest:
Rebased #732 from master to run the tests with a new `yarn.lock` file.


gilest:
@gossi I had to update `ember-template-lint` to v4 in both packages to get the tests to pass.

Not sure why as it all works fine locally.

Do we need to change the dependency installation/caching strategy in CI somehow to support monorepo?

gossi:
Great update of lint package. Have no idea why CI didn't run in the respective pkg only (although specified).

Go ahead and merge it when rdy. I'll be off until end of march.

PR #754, opened 2022-03-29
Correct typos in FileDropzone API docs

AddisonG:
onDragEnter was copy-pasted 3 times, and the example code had a fragmented sentence:

> Drag and drop images onto this area to upload them or


gilest:
Great catch – thank you!

Issue #650, opened 2022-01-19
Correct typings


gossi:
As types were introduced in #620 - I now see some mistakes and collect them here as I go to land a PR later to fix them:

- `UploadFile.upload()` - are the `options` optional ?
- `test-support` is not typed as of now, either convert to `ts` or provide a `.d.ts` file.

PR #757, opened 2022-04-01
Convert HttpRequest and upload() to typescript

gilest:
As part of #650 and to simplify the upcoming v2 build


GabrielCousin:
Hi @gilest! Huge thanks for the v5 release, we tested it recently and it works perfectly :) Any chance you could release this TS changes? It's the last blocker before we could use the new version <3
Thanks again!

GabrielCousin:
@gilest Thanks for the release! On one hand, it solves existing issues, on the other hand, it brings new TS errors 😢 I will do my best with opening a PR shortly, if that sounds good to you.

gilest:
That's a shame. Would welcome a PR or issue detailing the problems you're having

Issue #764, opened 2022-04-11
`ember-modifier` v3.2 deprecations


bertdeblock:
`ember-modifier` v3.2 introduced new deprecations which will also be triggered by the modifiers in this addon.
The migration guide can be found [here](https://github.com/ember-modifier/ember-modifier/blob/master/MIGRATIONS.md).

PR #767, opened 2022-04-14
Drop ember modifier v2 and resolve deprecations

gilest:
#764

### DIFF
--- a/ember-file-upload/package.json
+++ b/ember-file-upload/package.json
@@ -59,7 +59,7 @@
     "ember-cli-babel": "^7.26.6",
     "ember-cli-htmlbars": "^5.7.1 || ^6.0.1",
     "ember-cli-typescript": "^5.0.0",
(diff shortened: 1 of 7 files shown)


ember-file-upload/package.json
```diff
@@ -59,7 +59,7 @@
     "ember-cli-babel": "^7.26.6",
     "ember-cli-htmlbars": "^5.7.1 || ^6.0.1",
     "ember-cli-typescript": "^5.0.0",
-    "ember-modifier": "^2.1.2 || ^3.0.0",
+    "ember-modifier": "^3.0.0",
```
bertdeblock, line 62: Probably safer to actually use `^3.2.7` here.
The new API's were only added in `3.2.0`:
https://github.com/ember-modifier/ember-modifier/blob/master/CHANGELOG.md#v320-2022-03-29

Issue #546, opened 2021-09-27
`File.upload()` should resolve with a `Reponse` object


jelhan:
### Current situation

`upload` method of `File` currently resolves with a custom object having `status`, `body` and `header` properties: https://github.com/adopted-ember-addons/ember-file-upload/blob/e7b4188f3d5cd5d9b1549fc845d1ad110155eda0/addon/system/http-request.js#L51-L55

To provide the parsed body, Ember File Upload contains code to parse the raw response body depending on the `Content-Type` response header: https://github.com/adopted-ember-addons/ember-file-upload/blob/e7b4188f3d5cd5d9b1549fc845d1ad110155eda0/addon/system/http-request.js#L37-L49

I was very surprised to see this today. I expected that `File.upload()` resolves with a [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response) object. I noticed that it is not when trying to use `ok` property and noticed that it does not exist.

### Proposal

I like to propose that `File.upload()` resolves with a `Response` object instead. This would have multiple benefits:

1. `Response` object provides a more ergonomic API than the custom solution we have currently.
2. Consumer do not need to learn about a custom object used only by Ember File Upload to represent a HTTP response.
3. We do not need to maintain the code for parsing the response body.

Consumers would need to parse the response body themselves. I think it is safe to assume that most servers response with JSON. `Response` object provides `json` method, which makes parsing response body as JSON very easy. This library currently supports HTML and XML additionally. Consumers relying on that parsing functionality may consider this change a drawback.

This would be a breaking change.

### Implementation notes

A `Response` object can be constructed manually like this:

```js
new Response(body, { status, statusText, headers });
```

The request is done using a good old [`XMLHttpRequest`](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest). All required information should be available on that object when the request has been processed. Constructing a `Response` using a `XMLHttpRequest` may look like this:

```js
new Response(XMLHttpRequest.response, {
  status: XMLHttpRequest.status,
  statusText: XMLHttpRequest.statusText,
  headers: parseHeaders(XMLHttpRequest.getAllResponseHeaders())
});
```

[...]

PR #770, opened 2022-04-17
`File.upload()` returns `Response`

gilest:
As requested in #546

Drops all implemented response parsers.

Ported header parsing logic to minimise change.


jelhan (approved):
Thanks a lot for picking this one up!

PR #773, opened 2022-04-25
Use TrackedMap for queues

kennstenicht:
To display all uploading files in queues it would be nice to make the Map a TrackedMap.

### DIFF
--- a/ember-file-upload/addon/services/file-queue.ts
+++ b/ember-file-upload/addon/services/file-queue.ts
@@ -4,6 +4,7 @@ import { registerDestructor } from '@ember/destroyable';
 import Queue from '../queue';
 import UploadFile from '../upload-file';
 import { QueueName } from 'ember-file-upload/interfaces';
+import { TrackedMap } from 'tracked-built-ins';
 
 export const DEFAULT_QUEUE = Symbol('DEFAULT_QUEUE');
 
@@ -17,7 +18,7 @@ export const DEFAULT_QUEUE = Symbol('DEFAULT_QUEUE');
  * the remaining uploads.
  */
 export default class FileQueueService extends Service {
-  queues: Map<QueueName, Queue> = new Map();
+  queues: TrackedMap<QueueName, Queue> = new TrackedMap();
 
   /**
    * Returns a queue with the given name


jelhan (changes requested):
Thanks a lot for opening this PR.

I think it should be considered a bug fix. `files` property is expected to be used in template. But it depends on state, which is not tracked.

Do you mind adding a test to prevent a regression?

kennstenicht:
I have looked into the failing tests and the problem is, that `this.fileQueue.findOrCreate` is called inside a getter. If the deprecated `@name` attribute is used instead of the `@queue` the value is assigned and used in the same computation. We could move the function to the constructor, but then we would lose the abbility to change the name dynamically. When I have some free time, i'll take a closer look at it.

jelhan:
> I have looked into the failing tests and the problem is, that `this.fileQueue.findOrCreate` is called inside a getter.

Do I assume correctly that you are talking about this line of code? https://github.com/adopted-ember-addons/ember-file-upload/blob/9390723b89d4961e7a415f3cd66b696b8a6ef282/ember-file-upload/addon/helpers/file-queue.ts#L45 It reads `queues` in this line: https://github.com/adopted-ember-addons/ember-file-upload/blob/9390723b89d4961e7a415f3cd66b696b8a6ef282/ember-file-upload/addon/services/file-queue.ts#L29 And afterwards mutates it in this file if it does not found a queue before: https://github.com/adopted-ember-addons/ember-file-upload/blob/9390723b89d4961e7a415f3cd66b696b8a6ef282/ember-file-upload/addon/services/file-queue.ts#L52 This is indeed expected to throw because mutating a value during render, which has been read before.

As a workaround we could introduce an untracked map to check existence of a queue without reading the tracked map of queues.

kennstenicht:
Sorry for the unclear message, i was talking about the getter inside the file-dropzone component.
https://github.com/adopted-ember-addons/ember-file-upload/blob/ae1040d9a8915f3a4545d8e7982d04c4eb524e7d/ember-file-upload/addon/components/file-dropzone.ts#L214-L220

nickschot:
> Thanks a lot for opening this PR.
> 
> I think it should be considered a bug fix. `files` property is expected to be used in template. But it depends on state, which is not tracked.
> 
> Do you mind adding a test to prevent a regression?

I think in that case the queues property itself should also be `@tracked` no? As the [flush method](https://github.com/adopted-ember-addons/ember-file-upload/blob/9390723b89d4961e7a415f3cd66b696b8a6ef282/ember-file-upload/addon/queue.ts#L196) mutates it.

gilest:
Hey folks. I will take a look at this soon.

Just a note for the meantime – it may be easier to make this change now since I've removed a lot of deprecated code in preparation for v6.

gilest:
Continued in #787

gilest:
@kennstenicht Fix has been released in 6.0.2

kennstenicht:
@gilest Thanks for your work, i will give it a try.

PR #787, opened 2022-05-21
Track `queues` map in `FileQueue` service

gilest:
As pointed out in #773

> To display all uploading files in queues it would be nice to make the Map a TrackedMap.

Since this property wasn't tracked some of the public API of the `FileQueue` service was not reactive.

Have implemented with the suggestion by @jelhan to

> introduce an untracked map to check existence of a queue without reading the tracked map of queues.

Includes a regression test which fails without using `TrackedMap`.


gilest:
@kennstenicht how does this look?

PR #792, opened 2022-05-25
Move .md files to package root

### LINKED ISSUE #791: Move README.md to package and create linkts for CHANGELOG.md and README.md at repository root
I noticed that the `CHANGELOG.md` is not at the root of the repository anymore. Instead it at `ember-file-upload/CHANGELOG.md`. This makes it difficult to discover the file and may break autodiscovery by tools like RenovateBot.

There seem to be different patterns regarding location of changelog file in monorepo:

- [ember-auto-import](https://github.com/ef4/ember-auto-import) has `CHANGELOG.md` in the package but additionally has a link at the repository root pointing to the `CHANGELOG.md` in the package.
- [tracked-toolbox](https://github.com/tracked-tools/tracked-toolbox) and [ember.js](https://github.com/emberjs/ember.js) have `CHANGELOG.md` at repository root and not in the package.
- [ember-resources](https://github.com/NullVoxPopuli/ember-resources) has `CHANGELOG.md` in package without a link at repository root. So the same as ember-file-upload currently.

For Ember Auto Import the `CHANGELOG.md` and other files have been moved explicitly to the package in https://github.com/ef4/ember-auto-import/pull/434 by @buschtoens. The argument given there that `README.md`, `CHANGELOG.md` and `LICENSE.md` should be part of the published package makes sense. I would recommend to use the same pattern.

I haven't noticed the issue with `README.md` before. But not having it in the package breaks documentation in package manager listing: https://www.npmjs.com/package/ember-file-upload So I think that one should be addressed as well.

gilest:
Ensures that readme, license and changelog are all included with the published package. Will also restore README display on Ember Observer and npm websites.

Adds symbolic links from repo root for GitHub rendering and discovery.

Fixes #791


buschtoens (approved):

jelhan (approved):
Thanks a lot for picking it up that quickly. 🚀

Issue #778, opened 2022-05-19
Getting file type as empty string in case of font files


21Aby:
While trying to upload some font files ( `font/otf` & `font/ttf` ), the uploaded file type is an `empty string`. This in-turn fails my S3 upload operation ( since the file doesn't have the supposed type )

Is this because of a missing support for the above mentioned types or any other particular issues ?

Also Is there a way to set the file type by myself ( Right now, Can't set directly, since the properties are defined as getter only )

Please update

gilest:
Hey @21Aby thank you for reporting.

I tried to reproduce this in Chrome with random fonts from my system but the type property worked as expected.

Could you please share:

- Version of `ember-file-upload` that you are using
- Name and version of browser you are using
- If possible, the specific font files you are testing
- If you want to be really helpful, a minimal reproduction in a public repo

21Aby:
@gilest Thank you for the response

- "ember-file-upload": "5.1.0"
- Chrome ( Version 101.0.4951.64 (Official Build) (x86_64) )
- I tried with the `.otf` & `.ttf` types from https://dl.dafont.com/dl/?f=mabook (Sorry, Attaching font file not supported )

![Screenshot 2022-05-20 at 1 50 36 PM](https://user-images.githubusercontent.com/37609856/169486494-c785dc12-4a4b-40e9-89d1-276bd80c2981.png)

gilest:
Strange. We only proxy the value of the native File object. 

I checked the type of both files using the addon itself and [on MDN's File.type page](https://developer.mozilla.org/en-US/docs/Web/API/File/type) and the types show up correctly for me.

Can you try the tool linked above and let me know your results.

<img width="795" alt="Screen Shot 2022-05-21 at 4 06 14 PM" src="https://user-images.githubusercontent.com/36919/169634805-5263e994-bbde-4d9d-8a51-687b2a6806b1.png">
<img width="800" alt="Screen Shot 2022-05-21 at 4 06 22 PM" src="https://user-images.githubusercontent.com/36919/169634807-5cc04c3d-0111-4d37-ac75-471bf2c0ab24.png">

According to MDN

[...]

PR #795, opened 2022-05-26
Implement `UploadFile#replaceWithType`

gilest:
Attempting to help with #778 where the `File.type` reported by browsers is unreliable and results in some types of files failing to upload.


gilest:
This API is a bit weird and the original issue has been resolved.

Issue #152, opened 2018-06-12
Mirage route handler leaves hanging promise on unrecognized image type


DLiblik:
Applies to `2.5.0`

If the mirage upload helper is used and a .tif file is uploaded (a valid image type but not loadable into an Image object), then the assigned mirage route is never reached because the upload helper function leaves a hanging promise.  This occurs because the image pipeline handler does not supply a handler for the reject side of the promise created to parse metadata out of the image - tifs can't be processed and so an error occurs, but the promise doesn't reject.

I don't have time to do a PR on this atm, but the fix is simple: handle onerror and return no metadata.

https://github.com/tim-evans/ember-file-upload/blob/2.5.0/addon/mirage/utils.js#L52

```js
  image(file, metadata) {
    return new RSVP.Promise(function (resolve, reject) {
      let img = new Image();
      img.onload = function () {
        resolve(img);
      };

// === new code ===
      img.onerror = function(e) {
        reject(e);
      }
// ===

      img.src = metadata.url;
    }).then(function (img) {
      return {
        width: img.naturalWidth,
        height: img.naturalHeight
      };

// === new code ===
    }).catch(function(err) {
      // "No metadata for you!"
      return {};
// ===

    });
  },
```

gilest:
Hopefully this has been fixed by #275

DLiblik:
Not really.  It solves the mechanical issue, but #275 now rejects on a valid image ("valid image" is not synonymous with "an image that can be rendered in whatever browser your code runs in" - rather it is that the image content is valid for the mime type).

A `.tiff` is quite valid, just not for browser display, hence my suggestion that no metadata be returned (or, even better, that a property be set on the metadata indicating that the image was successfully received but could not be parsed), vs. an error which would suggest that the image upload itself was unsuccessful - there is no solid way to tell the difference.  

Mirage's purpose is to proxy server behaviour, which IMHO does not include assessing the validity of images based on their support in an arbitrary browser and version (Chrome did support `.tiff` and then dropped it - this should not have caused a server proxy technology to switch from passing to failing).

DLiblik:
(And yes, I recognize that the code I provided also does not differentiate - the helper here ideally would be updated to do that check so that the code using the helper knows what happened, as it would "in a real scenario").

jelhan:
Can you still reproduce this bug on latest master? If so a reproduction would be helpful. Or even better: a failing test.

gilest:
[...]

PR #797, opened 2022-05-30
Mirage hander: Don't send error response when file metadata can't be generated

gilest:
Instead the boolean property `hasAdditionalMetadata` reports  whether additional metadata was extracted.

After checking #152 I agree with the issue reporter. Throwing an error response from our provided handler because metadata cannot be generated by the brower's environment is unhelpful to users.

Users of the handler are free to respond with any response codes and errors themselves via the handler's own API.

With the new `hasAdditionalMetadata` prop it would be trivial to reimplement the "respond with 500 if metadata extraction fails" functionality in user land.

Includes dumped files and metadata matchers to ensure all metadata pipelines are working.


jelhan (approved):
I like this change a lot. Needing to provide a valid file made writing tests unnecessary complex in many cases.

PR #798, opened 2022-05-31
Replace usage of ember array prototype .any() with native .some()

gilest:
Believe this may throw an exception for users who don't have prototype extensions enabled.


bertdeblock:
I would also disable prototype extensions in the test app:
https://github.com/adopted-ember-addons/ember-file-upload/blob/master/test-app/config/environment.js#L14-L17

bertdeblock:
The dummy app for v1 addons has `ember-disable-prototype-extensions` installed by default. You could install that addon in the test app as well or just disable them via the environment file. I also made this issue https://github.com/embroider-build/addon-blueprint/issues/16 for the v2 addon blueprint.

PR #800, opened 2022-06-01
Migrate addon to v2

gilest:
### Mirage handler

I have a ready to go branch downstream of this which extracts the mirage handler into a new `ember-file-upload-mirage` package.

Moving the mirage handlers to their own package will be a breaking change and will be included with this change as a major release.

### Test waiter

I think that because we import `UploadFile` within the test-app the test waiter declaration is evaluated twice – results in a console warning. To at least avoid this I've replaced `buildWaiter` with `waitForPromise` which is run inline rather than in module scope anyway.

### TypeScript interfaces

I've moved most ts interfaces into a dedicated interfaces file. This is because importing across different modules creates a bit of a wild import map and can result in circular dependencies. I was advised to try this on Discord and it's working well.

It's (almost) compiled away by rollup, but it does contain some enums which are imported in the test-app so it's currently a public export. We could fix this with my next suggestion anyway...

### Module map

Since we're releasing breaking changes here – this could be a chance to improve the ergonomics of module exports.

I'd be in favour of named exporting all non-ember modules from `ember-file-upload`. To be included as a breaking change in v7. Can do this in a follow-up PR.


ember-file-upload/package.json
```diff
@@ -33,79 +33,63 @@
...
-    "test:ember-compatibility": "ember try:each",
-    "postversion": "git push && git push --tags",
-    "prepack": "ember ts:precompile",
-    "postpack": "ember ts:clean"
+    "start": "rollup --config --watch",
+    "test": "echo 'Addon does not have tests, run tests in test-app'",
+    "prepare": "yarn build"
   },
   "dependencies": {
+    "@ember/test-helpers": "^2.4.2",
```
gilest, line 49: Not certain these need to be declared as a dependency.

They are imported and used in our test helpers.

ember-file-upload/src/helpers/file-queue.ts
```diff
@@ -36,9 +30,9 @@ interface FileQueueArgs {
  * ```
  */
 export default class FileQueueHelper extends Helper implements QueueListener {
-  @service declare fileQueue: FileQueueService;
+  @service fileQueue!: FileQueueService;
```
gilest, line 33: Replaced the declare keyword here because the default typescript transpiler complained about it 🤷🏻

knownasilya: Hum, I'd figure out why it's complaining. I think we want `declare`

gilest: Reverted to using `declare`. Configured with the following in babel config.
```
["@babel/plugin-transform-typescript", { "allowDeclareFields": true }],
```

PR #802, opened 2022-06-11
Extract mirage handler to `ember-file-upload-mirage` package

gilest:
Configure as v2 addon.

Also migrated its codebase to typescript.

And renamed export to `uploadHandler`.

### Concerns

In terms of version, release and changelog I have some concerns.

Seems like there are three options:

1️⃣  - Sync version of both packages. Release both packages with every version increment. Prefix or otherwise indicate in changelog entries which package changes are relevant to

2️⃣  - Independent version for each package. Seems like this would require manually managing changelogs...

3️⃣ - Creating a new repo for the new package... Don't really want to do this


gilest:
After discussing with some stakeholders we've decided this isn't necessary - at least for now.

Reasoning:
- Maintenance overhead of managing releases, changelogs, versioning
- With the v2 addon it seems like the handler will only be included in bundles if it's explicitly imported

> fwiw, I tested the master branch (v2 addon) in a classic build and the handler file is completely excluded as well if not imported.

I've extracted the mirage handler improvements from this PR into #804

PR #803, opened 2022-06-11
Named export all modules

gilest:
Updates the public API of all exported modules to be named exports from the top level.

Another breaking change but will give us more flexibility in the future to move files around etc.

### DIFF
--- a/ember-file-upload/src/index.ts
+++ b/ember-file-upload/src/index.ts
@@ -1,0 +1,22 @@
+import UploadFile from './upload-file';
+import Queue from './queue';
+import HTTPRequest from './system/http-request';
+import DataTransferWrapper from './system/data-transfer-wrapper';
+import UploadFileReader from './system/upload-file-reader';
+import { FileSource, FileState } from './interfaces';
+import { DEFAULT_QUEUE } from './services/file-queue';
+
+export {
+  // Public API classes
+  Queue,
+  UploadFile,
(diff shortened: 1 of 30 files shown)


ember-file-upload/src/index.ts
```diff
@@ -0,0 +1,22 @@
...
+import { DEFAULT_QUEUE } from './services/file-queue';
+
+export {
+  // Public API classes
+  Queue,
+  UploadFile,
+  // Undocumented classes
+  HTTPRequest,
+  DataTransferWrapper,
+  UploadFileReader,
```
jelhan, line 16: Any reasons to export these classes at all? If they are privat, maybe we don't need a public export from them.

jelhan (approved):
I like this change a lot. Makes teaching public API way easier if there is only one import path.

ember-file-upload/src/index.ts
gilest: They're used in the test app. Otherwise I don't have a reason

jelhan: Maybe we could export them from `ember-file-upload/internal`? The import path would highlight that they are not part of public API. And editors won't include them as suggestion for autocompletion.

gilest: Great suggestion – much more clarity. I have implemented this

PR #833, opened 2022-08-05
feat(test-support): allow selectFiles input to be an element or a css selector

gvdp:
Using the test support helpers in our own projects we had the need to have the input of this method be an element instead of a string selector, so I added the option.

### DIFF
--- a/ember-file-upload/src/test-support.ts
+++ b/ember-file-upload/src/test-support.ts
@@ -24,12 +24,15 @@ import { assert } from '@ember/debug';
   Returns `Promise<void>` which resolves when the application is settled.
 
   @function selectFiles
-  @param {String} selector DOM selector for a FileUpload input
+  @param {(String | HTMLElement)} selector DOM selector for a FileUpload input or the FileUpload input HTML Element
   @param {File} ...files One or more File objects
   @return {Promise}
  */
-export async function selectFiles(selector: string, ...files: (File | Blob)[]) {
-  const input = find(selector);
+export async function selectFiles(
+  selector: string | HTMLElement,
+  ...files: (File | Blob)[]
+) {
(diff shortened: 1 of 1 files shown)


gilest:
Hey @gvdp thanks for opening this 👋🏻 

I'd be happy to accept this enhancement with a few additions:
- make the same change to all test helper functions exported from this file
- update the JSDoc comments for all test helpers also

gvdp:
Thanks for the feedback @gilest ! 

I implemented the change for all the methods in the test-support file. Not super familiar with JSDoc, I hope I did that right. Should I maybe also add an example of the HTMLElement input option?

ember-file-upload/src/test-support.ts
```diff
@@ -24,12 +24,15 @@ import { assert } from '@ember/debug';
   Returns `Promise<void>` which resolves when the application is settled.
 
   @function selectFiles
-  @param {String} selector DOM selector for a FileUpload input
+  @param {(String | HTMLElement)} selector DOM selector for a FileUpload input or the FileUpload input HTML Element
```
gilest, line 27: A few notes:
- `string` type should be lowercase
- Maybe we should rename `selector` to `target`. This is the param name [used by ember's test helpers](https://github.com/emberjs/ember-test-helpers/blob/master/API.md#click). I think it would make the code itself read better, too

```suggestion
  @param {(string | HTMLElement)} target The element or selector representing a file input
```

gilest (commented):
Thanks again @gvdp for those improvements.

Would like to make a few more for readability. If you don't have time or inclination I'm happy to pick this up and finish it for you 😅

gilest:
Was finished in #840

gvdp:
> Was finished in #840

Great! Thanks for picking this up 👍

Issue #2, opened 2016-04-05
Investigate folder uploads


tim-evans:
This looks abandoned; it may be worthwhile to investigate briefly.

jelhan:
There are two non-standard APIs that allow directory uploads. Both are controlled by properties on `<input type="file">`:

- `allowdirs`
- `webkitdirectory`

I'm tending toward not adding official support before one of them (or a third option) has been standardized.

zenhack:
It looks like those attributes are for controlling whether directories can be submitted as input; not setting them does not prevent directories from being submitted. I came across this looking into a downstream bug (https://github.com/mnutt/davros/issues/140) where trying to drag & drop a directory just results in uploading an empty, regular file.

It seems like this could be implemented using the `webkitrelativepath` property on the `File` object, which is also non-standard, but it *is* very widely supported: https://caniuse.com/mdn-api_file_webkitrelativepath; I would argue it doesn't make sense to wait on the standards bodies for this.

gilest:
Thanks for your input @zenhack. Would you be interested in preparing a PR?

zenhack:
I think I'm unlikely to find time to work on it myself at least in the short to medium term. And I'm probably the wrong person for the job -- I've never even used ember myself; my interest is as an end-user of said downstream app.

oliverlj:
this react plugin can upload a folder recursively :
https://github.com/rpldy/react-uploady/tree/master/packages/ui/upload-drop-zone

PR #834, opened 2022-08-05
feat: add option to allow dropping a directory and reading the contents

gvdp:
This is an attempt at resolving https://github.com/adopted-ember-addons/ember-file-upload/issues/2 , to support dropping folders.

We needed this for one of our clients, so I tried to add the option. It's a bit convoluted but the whole File and Directories web API is not so easy to use. Definitely open for ideas (and suggestions for better testing / documentation), but it would be great if we could add this to the package, for now we're publishing and using a forked version which works but is not ideal.

It also does use the non-standard `webkitGetAsEntry` file api but as said in this comment it is widely supported:
https://github.com/adopted-ember-addons/ember-file-upload/issues/2#issuecomment-1082435129
https://caniuse.com/?search=webkitGetAsEntry

### DIFF
--- a/ember-file-upload/src/components/file-dropzone.ts
+++ b/ember-file-upload/src/components/file-dropzone.ts
@@ -65,6 +65,10 @@ export default class FileDropzoneComponent extends Component<FileDropzoneArgs> {
     return this.args.multiple ?? true;
   }
 
+  get allowFolderDrop() {
+    return this.args.allowFolderDrop ?? false;
+  }
+
   get files(): File[] {
     const files = this.dataTransferWrapper?.files ?? [];
     if (this.multiple) return files;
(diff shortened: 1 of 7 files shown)


gilest:
Thanks for this contribution. I'll try to review later this week

bertdeblock:
Hey @gilest, would you be willing to review, if we update this PR?

gilest:
> Hey @gilest, would you be willing to review, if we update this PR?

Yes I'm open to reviewing it. From memory it looked ok but there's a bit of async-waterfall happening.

Was curious about how in-demand this feature might be but it seems reasonable to include

gvdp:
I'll close this one since I recreated it based on the latest master changes so there are no more merge conflicts:
https://github.com/adopted-ember-addons/ember-file-upload/pull/1066

PR #880, opened 2022-12-20
Add `ember-source` as a devDependency of `ember-file-upload`

gilest:
As suggested in the release notes of the embroider upgrade in #869


gilest:
Eas taken care of in #894

PR #882, opened 2022-12-23
Upgrade ember-resolver to 8.1.0

gilest:
Final version with ember 3.x support.


gilest:
Will let renovate handle this one. Don't need to keep this dep back as it's only internal, anyway.

PR #888, opened 2023-01-07
support ember-modifier v4

jelhan:
This adds support for ember-modifier v4. ember-modifier v3 is supported in parallel. This eases upgrades for applications, which have an indirect dependency on ember-modifier through several packages like this one.

As ember-modifier v3 is still supported going forward, this is _not_ a breaking change.

### DIFF
--- a/ember-file-upload/src/interfaces.ts
+++ b/ember-file-upload/src/interfaces.ts
@@ -82,16 +82,6 @@ export enum FileSource {
   Blob = 'blob',
 }
 
-export interface SelectFileModifierSignature {
-  Args: {
-    Named: {
-      filter?: (file: File, files: File[], index: number) => boolean;
-      onFilesSelected?: (files: UploadFile[]) => void;
-    };
-  };
-  Element: HTMLInputElement;
-}
-
 export interface FileDropzoneArgs {
   queue?: Queue;
 
--- a/ember-file-upload/src/queue.ts
+++ b/ember-file-upload/src/queue.ts
@@ -230,7 +230,7 @@ export class Queue {
         element.removeEventListener('change', changeHandler);
       };
     },
-    // ember-modifier@^3 requires an options hash as second argument
+    // @ts-expect-error ember-modifier@^3 requires an options hash as second argument
     // used to opt-in to lazy argument handling, which is the default for ember-modifier@^4
     { eager: false }
   );
--- a/ember-file-upload/src/queue.ts
+++ b/ember-file-upload/src/queue.ts
@@ -3,12 +3,7 @@ import { modifier } from 'ember-modifier';
 import { TrackedSet } from 'tracked-built-ins';
 import { UploadFile } from './upload-file';
 import type FileQueueService from './services/file-queue';
-import {
-  FileSource,
-  FileState,
-  QueueListener,
-  QueueName,
-} from './interfaces';
+import { FileSource, FileState, QueueListener, QueueName } from './interfaces';
 
 /**
  * The Queue is a collection of files that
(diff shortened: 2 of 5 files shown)


ember-file-upload/src/interfaces.ts
```diff
@@ -82,16 +82,6 @@ export enum FileSource {
...
-export interface SelectFileModifierSignature {
-  Args: {
-    Named: {
-      filter?: (file: File, files: File[], index: number) => boolean;
-      onFilesSelected?: (files: UploadFile[]) => void;
-    };
-  };
-  Element: HTMLInputElement;
-}
-
```
jelhan, line 94: I faced a TypeScript error when trying to use an interface. Inlining fixed it. I have no idea why. Not sure if inlining is acceptable. I also asked on Discord about a better solution: https://discord.com/channels/480462759797063690/484421406659182603/1061264497924382770

ember-file-upload/src/queue.ts
```diff
@@ -226,6 +230,8 @@ export class Queue {
         element.removeEventListener('change', changeHandler);
       };
     },
+    // @ts-expect-error: ember-modifier@^3 requires an options hash as second argument
+    // used to opt-in to lazy argument handling, which is the default for ember-modifier@^4
     { eager: false }
```
jelhan, line 235: Not sure if using `@ts-expect-error` is the correct approach here. I covered that one in the Discord question as well.

jelhan (commented):
It's working. But I'm not sure if I handled the TypeScript issues the right way. I left comments on each of them.

ember-file-upload/src/interfaces.ts
gilest: I think it's fine since it's not imported anywhere else.

When we recently fixed `ember-modifier` deprecations we had to adjust the signature then, also.

ember-file-upload/src/queue.ts
gilest: I guess the other way would be to detect the `ember-modfier` version with embroider macros. As long as this actually works on both versions I think that's fine.

Just a thought – should we add ember-try scenarios to ensure both versions of `ember-modfiier` work?

gilest (approved):
Looks good to me. Left a few thoughts but nothing preventing a merge really.

Just a note – I don't really want to release until I resolve the supposed dependency hoisting issue with `@ember/test-helpers`. Otherwise we'll ship a restrictive version requirement of that dependency.

Trying to resolve that now – best option may be to switch to pnpm...

Reference: https://github.com/adopted-ember-addons/ember-file-upload/pull/885

ember-file-upload/src/queue.ts
jelhan: > I guess the other way would be to detect the `ember-modfier` version with embroider macros. As long as this actually works on both versions I think that's fine.

I actually tried that. TypeScript even complains if passing `undefined`. So we would need to duplicate entire function call. I'm not sure if that would be a good trade-off just for making TypeScript happy.

gilest: I agree 👍🏻

jelhan:
> don't really want to release until we resolve the supposed dependency hoisting issue with `@ember/test-helpers`. Otherwise we'll ship a restrictive version requirement of that dependency.

No hurry from my side. Just ping me when I should rebase.

gilest:
@jelhan Please go ahead and rebase 😄

jelhan (commented):
Seems that you have already rebased. Looks good to me.

Would love to officially approve it. But GitHub UI seems to be broken. I cannot select "approve" in the radio buttons. 🤷

gilest:
> Would love to officially approve it. But GitHub UI seems to be broken. I cannot select "approve" in the radio buttons. 🤷

I think maybe you can't approve it because it's originally your PR

PR #890, opened 2023-01-09
Add interfaces folder to addon.publicEntrypoints

BoussonKarel:
In our app, we use TypeScript and import the QueueName interface from 'ember-file-upload/interfaces'

Because it is not added to publicEntrypoints of the V2 addon, the file we import changes to 'ember-file-upload/interfaces-fc8a799a'

### DIFF
--- a/ember-file-upload/rollup.config.js
+++ b/ember-file-upload/rollup.config.js
@@ -16,6 +16,7 @@ export default defineConfig({
     addon.publicEntrypoints([
       'components/**/*.ts',
       'helpers/**/*.ts',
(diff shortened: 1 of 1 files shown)


gilest:
Hey – thanks for your PR!

Not all interfaces are public API so I don't want to export all of them. However we could export the `QueueName` type from the [addon index file](https://github.com/adopted-ember-addons/ember-file-upload/blob/master/ember-file-upload/src/index.ts). Would you like to PR this?

BoussonKarel:
Ok, please check #897 @gilest

PR #894, opened 2023-01-14
Switch to pnpm and restore `@ember/test-helpers` to latest patch

gilest:
Allows us to continue to test Ember 3.28 via ember try.

I've been struggling with that particular build since: https://github.com/emberjs/ember-test-helpers/issues/1232

To my best understanding a quirk in yarn's (incorrect) handling of peerDependencies with monorepo causes incorrect resolution of `ember-source`.

I wasn't able to find a `nohoist` config with yarn which worked, although I'm open to suggestions.

As [suggested in this comment](https://github.com/embroider-build/embroider/issues/1169#issuecomment-1156803926) pnpm handles peerDeps more correctly, and provides the [dependenciesMeta.*.injected](https://pnpm.io/package_json#dependenciesmetainjected) option.

This issue has been blocking the release of `ember-file-upload` because this addon includes `@ember/test-helpers` in its own dependencies (we ship custom test helpers which depend on that package). I don't want to cut a release with a restrictive `@ember/test-helpers` dependency version.


ember-file-upload/package.json
```diff
@@ -63,12 +63,16 @@
     "@babel/plugin-transform-runtime": "^7.18.2",
     "@babel/plugin-transform-typescript": "^7.18.4",
     "@babel/preset-typescript": "^7.17.12",
+    "@babel/runtime": "^7.20.7",
+    "@ember/string": "^3.0.1",
```
jelhan, line 67: This may need to go into `peerDependency` to avoid a deprecation in Ember 4.10+. Adding as peer dependency would be a breaking change. Maybe better done in separate PR.

When doing a new major for that, we should also revisit other dependencies and see if they should be peer dependencies instead. E.g. `@glimmer/tracking` and `@glimmer/component`.

jelhan (approved):
Looks good to me. But I'm not a pnpm expert.

ember-file-upload/package.json
gilest: Ah good point. Definitely better done in a separate PR!

gilest: Added a tracking issue #899

PR #895, opened 2023-01-18
Issue related to async await for upload 

olenderhub:
I noticed that there are missing `return` in some places.

I followed https://ember-file-upload.pages.dev/docs/aws

Which is:

```
  @task({ maxConcurrency: 3, enqueue: true })
  *uploadImage(file) {
    const { product } = this.args;
    const apiResponse = yield fetch('/api/s3_direct');
    const s3Response = yield file.upload(apiResponse.url, {
        data: apiResponse.credentials
      });
    product.url = s3Response.headers.Location;
    yield product.save();
  }
  ```
  
  And this doesn't work on `const s3Response` - not resolving without `return resolve`
  
  Also, in this doc, there should be 
  
  ```
   const apiResponse = yield fetch('/api/s3_direct');
   const apiResponseResult = yield apiResponse.json();
   ```
   
   the same with s3Response but e.g. `yield s3Response.text()` (there could be a parser for XML).
   
   I can update the documentation when this is accepted and merged.

### DIFF
--- a/ember-file-upload/src/system/http-request.ts
+++ b/ember-file-upload/src/system/http-request.ts
@@ -73,7 +73,7 @@ export default class HTTPRequest {
     };
     this.request.onabort = bind(this, function () {
       this.onabort?.();
-      aborted.resolve();
+      return aborted.resolve();
     });
 
     this.request.onloadstart =
@@ -90,14 +90,14 @@ export default class HTTPRequest {
     this.request.onload = bind(this, function () {
       const response = parseResponse(this.request);
       if (Math.floor(response.status / 200) === 1) {
-        resolve(response);
+        return resolve(response);
       } else {
-        reject(response);
+        return reject(response);
       }
     });
 
     this.request.onerror = bind(this, function () {
-      reject(parseResponse(this.request));
+      return reject(parseResponse(this.request));
     });
 
     Object.defineProperty(this, 'timeout', {
@@ -113,7 +113,7 @@ export default class HTTPRequest {
 
     this.request.ontimeout = bind(this, function () {
       this.ontimeout?.();
-      reject(parseResponse(this.request));
+      return reject(parseResponse(this.request));
     });
   }
 

### CODE REVIEW

**Comment by gilest** [comment]
Hey @olenderhub – thanks for opening this.

First of all regarding the S3 examples in the docs. These have been ported from many years ago – I'm not surprised that they don't work without adjustment. Will happily accept PRs to update those examples!

## Async issue

I'm a bit unsure what you're encountering.

Could you please create a simple reproduction app or add a failing test to demonstrate the issue.

For example if I run the tests and put a debugger [here](https://github.com/adopted-ember-addons/ember-file-upload/blob/master/test-app/app/components/demo-upload.js#L35).

The contents of the HTTP response is available (see console output in bottom right).

Screen Shot 2023-01-20 at 2 28 07 PM

**Comment by olenderhub** [comment]
@gilest thank you for your reply. 

I checked this in a different combination today, and it works for me. I swear I had an issue yesterday with this. Only adding these returns helped me, but for now, I see I don't need them anymore. Maybe I had some problems with the chrome browser, I don't know, and I am confused.

So my PR can be close.
Just documentation for S3, I guess could be:

```
    const apiResponse = yield fetch('/api/s3_direct');
    const s3Response = yield file.upload(apiResponse.url, {
        data: apiResponse.credentials
      });
    product.url = s3Response.headers.Location;
    yield product.save();
```

e.g.
-->

```
    const apiResponse = yield fetch('/api/s3_direct');
    const apiResponseJson = yield apiResponse.json();
    const s3Response = yield file.upload(apiResponseJson.url, {
        data: apiResponse.credentials
      });
    const s3ResponseText = yield s3Response.text();
    const parser = new DOMParser();
    const xmlDoc = parser.parseFromString(s3ResponseText,"text/xml");
    product.url =  xmlDoc.getElementsByTagName("Location")[0].textContent;
    yield product.save();
    ```

Please let me know if this is fine. Sorry for wrong PR.

gilest:
Thanks again for contributing. I've pulled the docs update into #898

PR #910, opened 2023-03-19
Add `onUploadStarted`, `onUploadSucceeded` and `onUploadFailed` to `file-queue` helper

### LINKED ISSUE #58: How best to handle on upload completion
My use-case is to make an API request after all files are uploaded and wonder what the best design pattern would be to handle this. 

I have found that it's necessary to decrement the files.length property of the fileQueue service. I can then test for it to be 0 and then make the final API call to in my case attach the completed files to a content entity.  Should I have the manually decrement the files.length property?

gilest:
Should fix #58

## Outstanding issues

- [x] (1) At the time that `onUploadSucceeded` and `onUploadFailed` run, the `UploadFile` will have a state of `FileState.Uploading`. Otherwise when state is updated they'll be removed from the queue (queue flushing is triggered by the `state =` setter). We can either document to explain this or make further changes to accommodate it.
  - See comment below for suggestion to address this
- [x] (2) Whether we recommend the user updates `file.state` within these callbacks – depends on how we resolve issue (1). If we do permit updates to `file.state` in these callbacks we'll need to make sure they aren't overwritten by library code
  - Based on discussion I'm hopeful that we won't need to support users setting `file.state` themselves.
  - See comment below for suggestion to address this
- [ ] (3) Update API docs

### DIFF
--- a/docs/uploading.md
+++ b/docs/uploading.md
@@ -15,17 +15,13 @@ Bind a [file input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/in
 {{/let}}
 ```
 
-After a file is added to your queue, the `onFileAdded` callback will be called. You may then upload the file to a URL by calling `file.upload()`. This returns a promise that is resolved when the file has finished uploading, or is rejected if the file couldn't be uploaded.
+After a file is added to your queue, the `onFileAdded` callback will be called. You may then upload the file to a URL by calling `file.upload()`. This returns a promise that is resolved when the file has finished uploading.
 
 ```js
 export default class ExampleComponent extends Component {
   @action
   async uploadPhoto(file) {
-    try {
-      const response = await file.upload('/api/images/upload');
-    } catch (error) {
-      file.state = 'aborted';
-    }
+    const response = await file.upload('/api/images/upload');
   }
 }
 ```
--- a/ember-file-upload/src/helpers/file-queue.ts
+++ b/ember-file-upload/src/helpers/file-queue.ts
@@ -54,4 +54,16 @@ export default class FileQueueHelper extends Helper implements QueueListener {
   onFileRemoved(file: UploadFile) {
     this.args.onFileRemoved?.(file);
   }
+
+  onUploadStarted(file: UploadFile) {
+    this.args.onUploadStarted?.(file);
+  }
+
+  onUploadSucceeded(file: UploadFile, response: Response) {
+    this.args.onUploadSucceeded?.(file, response);
+  }
+
+  onUploadFailed(file: UploadFile, response: Response) {
+    this.args.onUploadFailed?.(file, response);
+  }
 }
--- a/ember-file-upload/src/system/upload.ts
+++ b/ember-file-upload/src/system/upload.ts
@@ -105,15 +105,22 @@ export function upload(
   file.state = FileState.Uploading;
 
   return waitForPromise(
-    uploadFn(request, options).then(
-      function (response) {
-        file.state = FileState.Uploaded;
+    uploadFn(request, options)
+      .then(function (response) {
+        file.queue?.uploadSucceeded(file, response);
+        // Don't update file.state if it was updated in the callback
+        if (file.state === FileState.Uploading) {
+          file.state = FileState.Uploaded;
+        }
         return response;
-      },
-      function (error) {
-        file.state = FileState.Failed;
-        return RSVP.reject(error);
-      }
-    )
+      })
+      .catch(function (response) {
+        file.queue?.uploadFailed(file, response);
+        // Don't update file.state if it was updated in the callback
+        if (file.state === FileState.Uploading) {
+          file.state = FileState.Failed;
+        }
+        return RSVP.reject(response);
+      })
   );
 }
--- a/ember-file-upload/src/system/upload.ts
+++ b/ember-file-upload/src/system/upload.ts
@@ -98,29 +98,26 @@ export function upload(
 
   request.ontimeout = () => {
     file.state = FileState.TimedOut;
+    file.queue?.flush();
   };
   request.onabort = () => {
     file.state = FileState.Aborted;
+    file.queue?.flush();
   };
   file.state = FileState.Uploading;
 
   return waitForPromise(
     uploadFn(request, options)
       .then(function (response) {
+        file.state = FileState.Uploaded;
         file.queue?.uploadSucceeded(file, response);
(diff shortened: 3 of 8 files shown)


test-app/tests/helpers/file-queue-helper-test.js
```diff
@@ -110,6 +113,90 @@ module('Integration | Helper | file-queue', function (hooks) {
...
+          Select File
+        </label>
+      {{/let}}
+    `);
+
+    await selectFiles('input[type=file]', new File([], 'dingus.txt'));
+
+    assert.verifySteps([
+      'upload succeeded: dingus.txt, state: uploading, response.status: 201',
+    ]);
```
gilest, line 167: As explained in (1) the state of this file is `FileState.Uploading` which is kind of unexpected due to this being the `onUploadSucceeded` callback.

ember-file-upload/src/system/upload.ts
```diff
@@ -105,15 +105,22 @@ export function upload(
...
-    uploadFn(request, options).then(
-      function (response) {
-        file.state = FileState.Uploaded;
+    uploadFn(request, options)
+      .then(function (response) {
+        file.queue?.uploadSucceeded(file, response);
+        // Don't update file.state if it was updated in the callback
+        if (file.state === FileState.Uploading) {
+          file.state = FileState.Uploaded;
+        }
```
gilest, line 114: Regarding (2)

test-app/tests/helpers/file-queue-helper-test.js
jelhan: I would tend towards fixing that if possible. It may be very confusing to the user. And it makes some use cases more challenging. E.g. a user can not handle both settled states (`onUploadSucceeded`, `onUploadFailed`) with the same handler if needing to know at any time processing the files of was successful or not.

jelhan: > when state is updated they'll be removed from the queue 

Is it a problem if the file is removed from queue when `onUploadSucceeded` and `onUploadFailed` is called?

ember-file-upload/src/system/upload.ts
jelhan: Why would a user update the file state?

ember-file-upload/src/helpers/file-queue.ts
```diff
@@ -54,4 +54,16 @@ export default class FileQueueHelper extends Helper implements QueueListener {
...
+
+  onUploadStarted(file: UploadFile) {
+    this.args.onUploadStarted?.(file);
+  }
+
+  onUploadSucceeded(file: UploadFile, response: Response) {
+    this.args.onUploadSucceeded?.(file, response);
+  }
+
+  onUploadFailed(file: UploadFile, response: Response) {
```
jelhan, line 66: Could `onUploadFailed` being used to implement an automatic retry?

test-app/tests/helpers/file-queue-helper-test.js
```diff
@@ -110,6 +113,90 @@ module('Integration | Helper | file-queue', function (hooks) {
...
+  test('will be notified when an upload starts', async function (assert) {
+    this.server.post(
+      '/upload-file',
+      uploadHandler(() => {})
+    );
+
+    this.upload = (file) => file.upload('/upload-file');
+
+    this.uploadStarted = (file) =>
+      assert.step(`upload started: ${file.name}, state: ${file.state}`);
```
jelhan, line 125: Test is easier to read if you split it up into two assertions in my opinion. One checking that the callback is executed. One checking that file state is as expected.

```suggestion
      assert.step('upload started');
      assert.propContains(file, { name: dingus.txt, state: uploading });
```

jelhan (commented):
I like it. Thanks a lot for working on it.

We ma see feature requests to add event when all files in the queue have been uploaded. But that we could add later.

Another important timing question is related to `Queue.files`. Is the file removed from the queue before or after? Should define, document and add test coverage for that as well.

test-app/tests/helpers/file-queue-helper-test.js
gilest: > I would tend towards fixing that if possible. It may be very confusing to the user

Agreed.

> Is it a problem if the file is removed from queue when onUploadSucceeded and onUploadFailed is called?

No I don't think that would cause any problems.

ember-file-upload/src/system/upload.ts
gilest: Good point. I suppose I just guessed that they might for some unknown reason. Like how some users update the file name.

I think the `state` setter might be private as only the getter is documented public API.

test-app/tests/helpers/file-queue-helper-test.js
gilest: Cool. Not used `propContains` before – thanks for the tip!

gilest: Unfortunately it doesn't seem like `propContains` supports class properties. Have separated the assertions for you, though.

gilest:
@jelhan I have made some internal adjustments to accommodate the issues raised in (1).

## Queue flushes

Have adjusted this functionality to explicit flushing instead of implicit flushing from file state setter. This means files will have the expected state within these callbacks.

It's a semi-breaking change as it's possible users use or rely on this functionality themselves even thought it's supposed to be internal.

## Result of `file.upload()`

Since we now have `onUploadFailed` I don't think this promise needs to reject any longer. I'm hoping this will fulfil the needs of #914 . This is a breaking change.

docs/uploading.md
```diff
@@ -15,17 +15,13 @@ Bind a [file input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/in
 {{/let}}
 ```
 
-After a file is added to your queue, the `onFileAdded` callback will be called. You may then upload the file to a URL by calling `file.upload()`. This returns a promise that is resolved when the file has finished uploading, or is rejected if the file couldn't be uploaded.
+After a file is added to your queue, the `onFileAdded` callback will be called. You may then upload the file to a URL by calling `file.upload()`. This returns a promise that is resolved when the file has finished uploading.
```
jelhan, line 18: Is the promise not rejecting anymore at all? Is it resolving even if upload fails? I'm confused by removing that part from the docs.

gilest: I got a bit carried away here – you can disregard this.

gilest:
Going to re-stage with a clearer explanation and docs changes.

PR #914, opened 2023-04-03
Return `Reponse.error` when status is outside of 200-599 range

herzzanu:
Currently we're creating a new [Response](https://github.com/adopted-ember-addons/ember-file-upload/blob/635e3350853e965c5df6a0065a85802b69f5c5e5/ember-file-upload/src/system/http-request.ts#L18-L25) for any response status. However, if the response status is outside of the `200...599` range, this will return an error. See https://developer.mozilla.org/en-US/docs/Web/API/Response for details on the accepted statuses. Also see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

The reasons for having a status outside of this range could be multiple, like a timeout issue or the server returning a status that does not necessarily comply with common accepted  statuses.

In this PR we're solving this issue by returning a custom Error when the status is outside of this range. The reason we return a custom error is that we want to give users the possibility to consume this error in their apps.

Next steps from here would be to differentiate these exceptions and return custom errors for each, similarly to what Ember Data does. Could be custom errors like `TimeoutError` for timeout for instance. This PR is an improvement nonetheless and can act as a good staring point. Any future improvements in handling these errors could be done independently, outside of this PR.

### DIFF
--- a/ember-file-upload/src/system/http-request.ts
+++ b/ember-file-upload/src/system/http-request.ts
@@ -17,11 +17,15 @@ function parseHeaders(headerString: string) {
 
 function parseResponse(request: XMLHttpRequest): Response {
   const body = request.response === '' ? null : request.response;
-  return new Response(body, {
-    status: request.status,
-    statusText: request.statusText,
-    headers: parseHeaders(request.getAllResponseHeaders()),
-  });
+  if (request.status >= 200 && request.status < 600) {
+    return new Response(body, {
+      status: request.status,
+      statusText: request.statusText,
+      headers: parseHeaders(request.getAllResponseHeaders()),
+    });
+  } else {
+    return Response.error();
+  }
 }
 
 export default class HTTPRequest {
--- a/test-app/tests/unit/system/http-request-test.js
+++ b/test-app/tests/unit/system/http-request-test.js
@@ -245,6 +245,20 @@ module('Unit | HttpRequest', function (hooks) {
     return promise;
   });
 
+  test('it returns an error response if the response status is ouside of the [200, 599] range', function (assert) {
+    assert.expect(3);
+
+    this.subject.open('PUT', 'http://emberjs.com');
+
+    this.subject.send({ filename: 'rfc.md' }).catch((response) => {
+      assert.strictEqual(response.status, 0, 'response status is 0');
+      assert.false(response.ok, 'response is not ok');
+      assert.strictEqual(response.type, 'error', 'response type is error');
+    });
+
+    this.respond(0, {}, 'timeout');
+  });
+
   skip('onprogress', function () {});
 
   skip('ontimeout', function () {});


jelhan:
Good catch!

> In this PR we're solving this issue by returning a custom Error when the status is outside of this range. The reason we return a custom error is that we want to give users the possibility to consume this error in their apps.
> 
> Next steps from here would be to differentiate these exceptions and return custom errors for each, similarly to what Ember Data does. Could be custom errors like `TimeoutError` for timeout for instance.

I would tend toward following the `fetch()` API and the exceptions thrown by those: https://developer.mozilla.org/en-US/docs/Web/API/fetch#exceptions Doing so would simplify both implementation and documentation. Especially we wouldn't need to invent our own exceptions but could use the existing once.

gilest:
Hi @herzzanu thank you very much for reporting this and providing a suggestion.

I tend to agree with @jelhan that I'd like to use native web APIs and minimise adding abstractions to this repo. (I've spent a long time removing similar abstractions over HTTP responses).

We're actually working on solving (hopefully) this same problem by providing more callbacks and flexibility to addon users.

In #910 I've removed the error response from `file.upload()`. Instead users can decide how to handle "failed" responses via the new callbacks. Would this be enough to satisfy your needs?

herzzanu:
@gilest Thanks for looking into it and for the work on #910 

I don't see how that one mitigates the issue where the response status is not in the [200, 500] range. Am I missing something?

herzzanu:
> I would tend toward following the `fetch()` API and the exceptions thrown by those: https://developer.mozilla.org/en-US/docs/Web/API/fetch#exceptions Doing so would simplify both implementation and documentation. Especially we wouldn't need to invent our own exceptions but could use the existing once.

@jelhan I'm all in for using a native approach here. I was surprised to see we're not using fetch.

jelhan:
If I recall correctly, the main blocker for using `fetch` is that tracking upload progress is not yet possible using `fetch`. It may be possible in the future. [Chrome recently shipped support for streaming request bodies](https://developer.chrome.com/articles/fetch-streaming-requests/). But it's [not available in Firefox and Safari](https://developer.mozilla.org/en-US/docs/Web/API/Request#browser_compatibility) yet. It's clear that we should go that path long-term in my opinion. That's why we are already aligning the public API with `fetch` as much as possible.

jelhan:
> We're actually working on solving (hopefully) this same problem by providing more callbacks and flexibility to addon users.
> 
> In #910 I've removed the error response from `file.upload()`. Instead users can decide how to handle "failed" responses via the new callbacks. Would this be enough to satisfy your needs?

I think the issue here is slightly different. The problem is that the `status` may be out of range causing the browser to throw when we construct the `Response`.

If I understand the issue description correctly, the following is happening in some cases:

```js
new Response('', {
  status: 0,
});
```

Browser throws an exception because `0` is not a valid HTTP status:

> Uncaught RangeError: Response constructor: Invalid response status code.

What I don't fully understand yet, in which cases the `load` event of `XMLHttpRequest` is fired but response has an invalid status code. Per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest?retiredLocale=de#events) it should the `load` event should only be fired when "transaction completes successfully".

@herzzanu you mentioned in PR description that network timeout could cause this. Are you sure that browser fires `load` event if the network connection times out? I would expect it fires `error` event in that case.

As the second case you are mentioning that a "server returning a status that does not necessarily comply with common accepted statuses." Did you investigated how `fetch` handles such situations? Does it throw? What status code has the `Response` in that case?

herzzanu:
> > We're actually working on solving (hopefully) this same problem by providing more callbacks and flexibility to addon users.
> > In #910 I've removed the error response from `file.upload()`. Instead users can decide how to handle "failed" responses via the new callbacks. Would this be enough to satisfy your needs?
> 
> I think the issue here is slightly different. The problem is that the `status` may be out of range causing the browser to throw when we construct the `Response`.
> 
> If I understand the issue description correctly, the following is happening in some cases:
> 
> ```js
> new Response('', {
>   status: 0,
> });
> ```
> 
> Browser throws an exception because `0` is not a valid HTTP status:
> 
> > Uncaught RangeError: Response constructor: Invalid response status code.
> 
> What I don't fully understand yet, in which cases the `load` event of `XMLHttpRequest` is fired but response has an invalid status code. Per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest?retiredLocale=de#events) it should the `load` event should only be fired when "transaction completes successfully".
> 
> @herzzanu you mentioned in PR description that network timeout could cause this. Are you sure that browser fires `load` event if the network connection times out? I would expect it fires `error` event in that case.
> 
> As the second case you are mentioning that a "server returning a status that does not necessarily comply with common accepted statuses." Did you investigated how `fetch` handles such situations? Does it throw? What status code has the `Response` in that case?

@jelhan yes, the error happens in some cases. We catch these cases in Sentry. I assumes it's a timeout since the response status is `0`, but can also be various reasons for that. Not sure it matters the actual reason, the issue is that we hit the part of the code where we try to create the `Response` with a code status equal to `0`.

gilest:
Apologies @herzzanu. Having tried this locally I believe this is a bug. It exposes some other untested features that probably don't work (or have been broken by changes) such as request timeouts (on the client side).

Anyway I found some time to research the native `fetch` APIs and found something we could consider using.

The static method `Response.error()` returns a generic response which represents a failed network request. [MDN ref](https://developer.mozilla.org/en-US/docs/Web/API/Response/error)

We can't set any properties on this object but since it represents a failure at the network level (presumably) that might be an ok tradeoff.

Seems to fulfil basically what we'd want here:

```js
    assert.expect(3);

    this.subject.open('PUT', 'http://emberjs.com');

    this.subject.send({ filename: 'rfc.md' }).catch((response) => {
      assert.strictEqual(response.status, 0, 'response status is 0');
      assert.strictEqual(response.ok, false, 'response is not ok');
      assert.strictEqual(response.type, 'error', 'response type is error');
    });

    this.respond(0, {}, 'timeout');
```

Let me know what you think @herzzanu @jelhan

jelhan:
@gilest `Response.error()` sounds perfect on case if network errors. Good catch!

I think I found the root cause of our issue. We are trying to parse the response even in `error` event handler: https://github.com/adopted-ember-addons/ember-file-upload/blob/78c2a1506f78694cbfe092f16fe9ff490b64d5c2/ember-file-upload/src/system/http-request.ts#L99-L101

Haven't looked into spec yet in which cases `error` event is thrown and if it's the same as fetch rejecting. But likely we want to reject with `Response.error()` in most cases.

herzzanu:
@gilest nice stuff. I think this will mitigate any other errors with statuses that don't fit the Response required range 👏

herzzanu:
Sounds good @gilest  I can make some time and update this PR

gilest (approved):
Looks perfect. Thanks 🙌🏻

gilest:
>  I think I found the root cause of our issue. We are trying to parse the response even in error event handler

This should be mitigated by the latest changes in the case of both `onerror` and `ontimeout`. We'll need to add test coverage for this if possible. It's non-trivial to mock though and should be outside of the scope of this bugfix.

gilest:
Released in [7.3.1](https://github.com/adopted-ember-addons/ember-file-upload/releases/tag/v7.3.1)

PR #917, opened 2023-04-07
Move blueprint provided packages to peerDependencies

### LINKED ISSUE #899: Audit peerDependencies
Pulled out of #894 where `devDependencies` have been added to ~~`@babel/runtime` and~~ `@ember/string`.

As suggested by @jelhan 

>... may need to go into peerDependency to avoid a deprecation in Ember 4.10+. Adding as peer dependency would be a breaking change. Maybe better done in separate PR.

> When doing a new major for that, we should also revisit other dependencies and see if they should be peer dependencies instead. E.g. @glimmer/tracking and @glimmer/component.

gilest:
Experiment to see how we might fix #899

Being aggressive here and moving as much as possible from `dependencies` to `devDependencies` in an attempt to batch breaking changes and avoid another major.

### DIFF
--- a/ember-file-upload/package.json
+++ b/ember-file-upload/package.json
@@ -46,15 +46,10 @@
     "prepare": "pnpm build"
   },
   "dependencies": {
-    "@ember/test-helpers": "^2.9.3",
     "@ember/test-waiters": "^3.0.0",
     "@embroider/addon-shim": "^1.5.0",
     "@embroider/macros": "^1.0.0",
-    "@glimmer/component": "^1.0.4",
-    "@glimmer/tracking": "^1.0.4",
-    "ember-auto-import": "^2.0.0",
-    "ember-modifier": "^3.2.7 || ^4.0.0",
-    "tracked-built-ins": "^3.0.0"
+    "ember-auto-import": "^2.0.0"
   },
   "devDependencies": {
     "@babel/core": "^7.21.4",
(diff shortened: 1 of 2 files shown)


ember-file-upload/package.json
```diff
@@ -99,12 +98,18 @@
...
   "peerDependencies": {
+    "@ember/test-helpers": "^2.9.3",
+    "@glimmer/component": "^1.1.2",
+    "@glimmer/tracking": "^1.1.2",
     "ember-cli-mirage": "*",
-    "miragejs": "*"
+    "ember-modifier": "^3.2.7 || ^4.1.0",
+    "miragejs": "*",
+    "tracked-built-ins": "^3.1.1"
   },
```
gilest, line 113: Everything I've moved to `peerDependencies` is in the modern `ember-cli` blueprints.

ember-file-upload/package.json
```diff
@@ -46,15 +46,10 @@
...
     "@ember/test-waiters": "^3.0.0",
     "@embroider/addon-shim": "^1.5.0",
     "@embroider/macros": "^1.0.0",
-    "@glimmer/component": "^1.0.4",
-    "@glimmer/tracking": "^1.0.4",
-    "ember-auto-import": "^2.0.0",
-    "ember-modifier": "^3.2.7 || ^4.0.0",
-    "tracked-built-ins": "^3.0.0"
+    "ember-auto-import": "^2.0.0"
   },
```
gilest, line 53: Believe the remaining dependencies are required. Keen for others to check this, though.

jelhan: Looks good to.

jelhan (approved):
Looks good to me.

PR #925, opened 2023-04-29
Add `onUploadStarted`, `onUploadSucceeded` and `onUploadFailed` to `file-queue` helper 

gilest:
Re-staged version of #910  since that PR was getting messy. Should be easier to review this one.

Adds callbacks to `file-queue` helper:
- `onUploadStarted`
- `onUploadSucceeded`
- `onUploadFailed`

So that `UploadFile` instances have a sensible state at callback time I have adjusted the queue flush from a side-effect of the `File.state` setter and instead call it explicitly. As I understand this is an internal change, only.

Should fix https://github.com/adopted-ember-addons/ember-file-upload/issues/58

### DIFF
--- a/docs/api/helper/file-queue.md
+++ b/docs/api/helper/file-queue.md
@@ -26,11 +26,14 @@ If a name is not provided, the default queue will be used.
 
 ## Arguments
 
-| Argument        | Description                       | Type                         | Default value |
-| --------------- | --------------------------------- | ---------------------------- | ------------- |
-| `name`          | Name of queue to find or create.  | `string`                     |               |
-| `onFileAdded`   | Called when each file is added.   | `(file: UploadFile) => void` |               |
-| `onFileRemoved` | Called when each file is removed. | `(file: UploadFile) => void` |               |
+| Argument            | Description                               | Type                                             | Default value |
+| ------------------- | ----------------------------------------- | ------------------------------------------------ | ------------- |
+| `name`              | Name of queue to find or create.          | `string`                                         |               |
+| `onFileAdded`       | Called when each file is added.           | `(file: UploadFile) => void`                     |               |
+| `onFileRemoved`     | Called when each file is removed.         | `(file: UploadFile) => void`                     |               |
+| `onUploadStarted`   | Called when each upload begins.           | `(file: UploadFile) => void`                     |               |
+| `onUploadSucceeded` | Called when each upload request succeeds. | `(file: UploadFile, response: Response) => void` |               |
+| `onUploadFailed`    | Called when each upload request fails.    | `(file: UploadFile, response: Response) => void` |               |
 
 ## Properties
 
--- a/docs/api/helper/file-queue.md
+++ b/docs/api/helper/file-queue.md
@@ -26,14 +26,14 @@ If a name is not provided, the default queue will be used.
 
 ## Arguments
 
-| Argument            | Description                               | Type                                             | Default value |
-| ------------------- | ----------------------------------------- | ------------------------------------------------ | ------------- |
-| `name`              | Name of queue to find or create.          | `string`                                         |               |
-| `onFileAdded`       | Called when each file is added.           | `(file: UploadFile) => void`                     |               |
-| `onFileRemoved`     | Called when each file is removed.         | `(file: UploadFile) => void`                     |               |
-| `onUploadStarted`   | Called when each upload begins.           | `(file: UploadFile) => void`                     |               |
-| `onUploadSucceeded` | Called when each upload request succeeds. | `(file: UploadFile, response: Response) => void` |               |
-| `onUploadFailed`    | Called when each upload request fails.    | `(file: UploadFile, response: Response) => void` |               |
+| Argument            | Description                                             | Type                                             | Default value |
+| ------------------- | ------------------------------------------------------- | ------------------------------------------------ | ------------- |
+| `name`              | Name of queue to find or create.                        | `string`                                         |               |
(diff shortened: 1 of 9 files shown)


docs/uploading.md
```diff
@@ -15,16 +15,18 @@ Bind a [file input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/in
...
   async uploadPhoto(file) {
     try {
       const response = await file.upload('/api/images/upload');
-    } catch (error) {
-      file.state = 'aborted';
+    } catch (response) {
+      // Handle error response somehow
     }
   }
 }
```
gilest, line 32: @jelhan as you say I don't think there's a reason consumers should update state. So I'm suggesting we replace this example

jelhan: I feel that make sense. Maybe we could give a more realistic example? I guess informing the user and providing an option to retry the upload might be most common error handling?

Something like this:

```js
export default class ExampleComponent extends Component {
  @action
  async uploadPhoto(file) {
    try {
      const response = await file.upload('/api/images/upload');
    } catch (response) {
      if (!response.ok) {
        const shouldRestart = window.confirm(
          'Uploading failed due to broken network connectivity. Please check your internet connection. Do you want to retry?'
        );
        
        if (shouldRestart) {
          await this.uploadPhoto(file);
        }
      } else {
        window.alert('Upload failed with a server-side error. Please try again later.');
      }
    }
  }
}
```

Or do we want to push the people towards `onUploadFailed` for error handling?

docs/api/helper/file-queue.md
```diff
@@ -26,11 +26,14 @@ If a name is not provided, the default queue will be used.
...
-| `onFileRemoved` | Called when each file is removed. | `(file: UploadFile) => void` |               |
+| Argument            | Description                               | Type                                             | Default value |
+| ------------------- | ----------------------------------------- | ------------------------------------------------ | ------------- |
+| `name`              | Name of queue to find or create.          | `string`                                         |               |
+| `onFileAdded`       | Called when each file is added.           | `(file: UploadFile) => void`                     |               |
+| `onFileRemoved`     | Called when each file is removed.         | `(file: UploadFile) => void`                     |               |
+| `onUploadStarted`   | Called when each upload begins.           | `(file: UploadFile) => void`                     |               |
+| `onUploadSucceeded` | Called when each upload request succeeds. | `(file: UploadFile, response: Response) => void` |               |
+| `onUploadFailed`    | Called when each upload request fails.    | `(file: UploadFile, response: Response) => void` |               |
```
jelhan, line 36: I find "each file" and "each upload" a little bit confusing wording. Maybe:

```suggestion
| `onFileAdded`       | Called when a file is added to this queue.           | `(file: UploadFile) => void`                     |               |
| `onFileRemoved`     | Called when a file is removed from this queue.         | `(file: UploadFile) => void`                     |               |
| `onUploadStarted`   | Called when upload starts for a file of this queue.           | `(file: UploadFile) => void`                     |               |
| `onUploadSucceeded` | Called when upload of a file of this queue succeeds. | `(file: UploadFile, response: Response) => void` |               |
| `onUploadFailed`    | Called when upload of a file of this queue fails fails.    | `(file: UploadFile, response: Response) => void` |               |
```

ember-file-upload/src/system/upload.ts
```diff
@@ -98,22 +98,26 @@ export function upload(
...
+      .catch(function (response) {
         file.state = FileState.Failed;
-        return RSVP.reject(error);
-      }
-    )
+        file.queue?.uploadFailed(file, response);
+        return RSVP.reject(response);
+      })
+      .finally(() => file.queue?.flush())
   );
```
jelhan, line 122: We may want rewriting this to `try {} catch {} finally {}` for readability at some time.

ember-file-upload/src/system/upload.ts
```diff
@@ -98,22 +98,26 @@ export function upload(
...
+      })
+      .catch(function (response) {
         file.state = FileState.Failed;
-        return RSVP.reject(error);
-      }
-    )
+        file.queue?.uploadFailed(file, response);
+        return RSVP.reject(response);
+      })
+      .finally(() => file.queue?.flush())
```
jelhan, line 121: I just noticed that `Queue.flush()` is a confusing naming. It sounds as if all files are removed from the queue. Not only those, which have been uploaded successfully or aborted. I was very confused what this code does until reading the API docs for `Queue.flush()`.

jelhan (approved):
Looks good to me. Comments are only meant to be proposals. No required changes from my side before merging.

docs/api/helper/file-queue.md
gilest: Thanks! I committed this with a few grammatical adjustments.

ember-file-upload/src/system/upload.ts
gilest: Agreed!

ember-file-upload/src/system/upload.ts
gilest: This is a fair point. The queue flushing has confused me a few times now 😅

docs/uploading.md
gilest: I think an example like this is good starting point but I'm very conscious that many users will copy and paste example code directly into their app (I've seen it happen).

Given this, we'd want a well-considered example that's going to be best-practice for most users.

I think advanced usage like retries, s3 uploads etc could probably have their own "Recipes" pages on the doc site. We could pull some example code from old issue solutions, too.

PR #938, opened 2023-06-05
Name `@babel/plugin-proposal-object-rest-spread` as a dependency

gilest:
This module is apparently missing in Floating Dependency builds.

Adding this package results in the next missing package: `@babel/plugin-proposal-async-generator-functions`.

Seems to be a pnpm/CI misconfiguration or something as I cannot reproduce locally.


gilest:
Closed in favour of #940

PR #940, opened 2023-06-09
Restore floating dependency builds

gilest:
Continuation of #938 now that I've found a configuration which allows builds to succeed on CI.

Believe the underlying issue is with `rollup-plugin-ts` as explained here: babel/babel/issues/15670

Also mentioned in passing here: embroider-build/addon-blueprint/pull/127


.npmrc
```diff
@@ -1 +1,2 @@
 auto-install-peers=false
+resolution-mode=highest
```
gilest, line 2: This resolution mode makes more sense to me than the default (oldest). Another way would could handle this is much more strict versions in our `package.json` files.

PR #944, opened 2023-06-11
Migrate `test-app` to TypeScript. Add Typecheck job to CI

gilest:
Main purpose is to add coverage for TypeScript types which we are shipping but not really testing their consumption/use.

Secondary purpose is to prepare infrastructure for testing and shipping Glint types.

Some internal system tests remain untyped as they will be more difficult to type. I'm ok with this for now as they aren't public API. It may be easier to refactor than them.


test-app/app/components/demo-upload.js
```diff
@@ -22,7 +22,7 @@ export default class DemoUpload extends Component {
       file,
     });
 
-    this.args.files.pushObject(asset);
+    this.args.files.push(asset);
```
gilest, line 25: Replace EmberArray with TrackedArray in tests.

NullVoxPopuli (approved):

NullVoxPopuli:
One thing you could do as a followup is at a version matrix to your typecheck job, having a range of supported ts versions 🎉

PR #950, opened 2023-06-18
Relax peer dependency version for @ember/test-helpers

seanCodes:
Allow `@ember/test-helpers` v3 to be used with this addon. v3 was released a few weeks ago and doesn’t seem to have any changes impacting this addon ([release notes](https://github.com/emberjs/ember-test-helpers/releases/tag/v3.0.0)). I’ve been using it with `ember-file-upload` for a while now and haven’t had any issues, so recommend relaxing the peer requirements so npm won’t protest when trying to install the latest test-helpers version.

For now I’ve set the version range to `^2.9.3 || ^3.0.3` (since 3.0.3 seems to be the [earliest stable version](https://github.com/emberjs/ember-test-helpers/releases#hd-22950539)) but could relax even more (`>= 2.9.3`?) if that makes sense.

### DIFF
--- a/ember-file-upload/package.json
+++ b/ember-file-upload/package.json
@@ -110,7 +110,7 @@
     "webpack": "^5.74.0"
   },
   "peerDependencies": {
-    "@ember/test-helpers": "^2.9.3",
+    "@ember/test-helpers": "^2.9.3 || ^3.0.3",
     "@glimmer/component": "^1.1.2",
     "@glimmer/tracking": "^1.1.2",
     "ember-cli-mirage": "*",


gilest:
Thanks for the PR.

> could relax even more (>= 2.9.3?) if that makes sense.

Agreed, declaring peerDeps is a bit new for the community, but at present I feel like more lenient peerDep ranges will avoid a lot of update churn. 

If we change that we can do it in future, though.

gilest:
Thanks again @seanCodes . This has been released in `v8.1.0`. Please let us know if you encounter anything with the new version as there's been significant internal changes since last release.

seanCodes:
Thanks @gilest! I’ll try it and let you know.

PR #961, opened 2023-07-21
Use stable types from `ember-source`

gilest:
Also upgrades internal packages to ember-source 5.1

Considering this as an internal change since these are development types...


test-app/tests/integration/components/file-dropzone-test.ts
```diff
@@ -22,7 +22,9 @@ module('Integration | Component | FileDropzone', function (hooks) {
   setupRenderingTest(hooks);
 
   hooks.beforeEach(function (this: LocalTestContext) {
-    const fileQueueService = this.owner.lookup('service:file-queue');
+    const fileQueueService = this.owner.lookup(
+      'service:file-queue'
+    ) as FileQueueService;
```
gilest, line 27: Is there a way to configure this with a `Registry`. I tried and failed but I've not used TypeScript in a test-app before

PR #969, opened 2023-08-08
Use Glint types in test-app

gilest:
Glint is pretty sweet!


test-app/tests/helpers/file-queue-helper-test.ts
```diff
@@ -70,8 +73,9 @@ module('Integration | Helper | file-queue', function (hooks) {
   });
 
   test('falls back to default name', async function (this: LocalContext, assert) {
-    await render(hbs`
+    await render<LocalContext>(hbs`
       {{#let (file-queue) as |queue|}}
+        {{! @glint-ignore: https://github.com/typed-ember/glint/discussions/611 }}
         <output>{{queue.name}}</output>
```
gilest, line 79: These are the only checks I'm ignoring.

The story is that `QueueName` has a type of `string | symbol` but a `symbol` is not a variant of Glint's `ContentValue` type which represents values which can be directly rendered by Glimmer.

Discussion reference: typed-ember/glint/discussions/611

PR #1005, opened 2023-10-21
Regression coverage for incorrect `loaded` & `progress` stats

gilest:
An attempt to prove and model  #1001 with a failing regression test

image

Unfortunately to do so I've also had to modify some internals...

### Mirage upload handler 🌴 

Adjusted this route handler to more accurately model a real upload request.

It now fires `onloadstart`, `onprogress` and `onloadend` rather then `onprogress` only.

The `total` value of the `onloadend` `ProgressEvent` is intentionally set to an arbitrary value.

⚠️ I don't think this is a breaking change, and I'd be surprised if it affected users of the mirage handler as it's a specific implementation detail of something this addon handles for users. No guarantees, though.

### HTTPRequest 🌐 

Setup so that `onloadstart` and  `onloadend` may be called from the Mirage handler.

Shouldn't affect addon users.

### DIFF
--- a/ember-file-upload/src/system/http-request.ts
+++ b/ember-file-upload/src/system/http-request.ts
@@ -82,15 +82,20 @@ export default class HTTPRequest {
       aborted.resolve();
     });
 
-    this.request.onloadstart =
-      this.request.onprogress =
-      this.request.onloadend =
-        bind(this, function (evt) {
-          this.onprogress?.(evt);
-        });
+    this.request.onloadstart = bind(this, function (evt) {
+      this.onprogress?.(evt);
+    });
+    this.request.onprogress = bind(this, function (evt) {
+      this.onprogress?.(evt);
+    });
+    this.request.onloadend = bind(this, function (evt) {
+      this.onprogress?.(evt);
+    });
 
     if (this.request.upload) {
+      this.request.upload.onloadstart = this.request.onprogress;
       this.request.upload.onprogress = this.request.onprogress;
+      this.request.upload.onloadend = this.request.onprogress;
     }
 
     this.request.onload = bind(this, function () {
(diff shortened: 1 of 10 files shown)


ember-file-upload/src/system/http-request.ts
```diff
@@ -82,15 +82,20 @@ export default class HTTPRequest {
...
+    });
+    this.request.onloadend = bind(this, function (evt) {
+      this.onprogress?.(evt);
+    });
 
     if (this.request.upload) {
+      this.request.upload.onloadstart = this.request.onprogress;
       this.request.upload.onprogress = this.request.onprogress;
+      this.request.upload.onloadend = this.request.onprogress;
     }
```
gilest, line 99: Note these are still intentionally wired up to call `onprogress` only. In the actual bugfix they will each call their own method.

test-app/app/components/demo-upload.ts
```diff
@@ -33,8 +34,7 @@ export class Asset {
 }
 
 export default class DemoUpload extends Component<DemoUploadSignature> {
-  @action
-  async uploadProof(file: UploadFile) {
+  uploadProof = enqueueTask(async (file: UploadFile) => {
```
gilest, line 37: Adjusted to upload files one-by-one for this purpose of this regression test.

test-app/tests/integration/progress-test.ts
```diff
@@ -0,0 +1,108 @@
...
+  onComplete = () => {};
+  promise = new Promise<void>((resolve) => {
+    this.onComplete = () => {
+      this.isCompleted = true;
+      resolve();
+    };
+  });
+}
+
+module.skip('Integration | progress', function (hooks) {
```
gilest, line 29: Can be unskipped downstream in the bugfix PR at which point this should pass ✅

PR #1007, opened 2023-10-23
Incorrect file loaded reporting

mkszepp:
When the upload is completed, `request.onprogress` will be fired at the same time like `request.onloadend`.
As the upload is completed with `request.onprogress` the server returns the `Content-Length` and the browser takes this size... In my case the returned `Content-Length` is lower, because i don't return the image and so the progress bar is going back.

My debug console.logs:
grafik

To fix this behavier we should replace the `file.loaded` set in `request.onprogress` to:
`file.loaded = Math.max(evt.loaded, file.loaded);` instead of `file.loaded = evt.loaded;`

Comment copied from: https://github.com/adopted-ember-addons/ember-file-upload/pull/1002#issuecomment-1775101785

### DIFF
--- a/ember-file-upload/src/system/upload.ts
+++ b/ember-file-upload/src/system/upload.ts
@@ -100,7 +100,7 @@ export function upload(
     if (!evt) return;
     if (!evt.lengthComputable || evt.total === 0) return;
 
-    file.loaded = evt.loaded;
+    file.loaded = Math.max(evt.loaded, file.loaded);
     file.progress = (file.loaded / file.size) * 100;
   };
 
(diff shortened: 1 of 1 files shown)


mkszepp:
@gilest one think, that we should discuss.. shouid we check in the "onprogress" if `file.loaded` is not > than file.size, otherwise it can occures more than 100% with this change

mkszepp:
@gilest friendly ping

gilest:
Hey @mkszepp thanks. As before I think this looks more correct than the previous implementation.

It would be cool to take your collected logs and use them to build a regression test but I don't think that should block this being released.

gilest:
Released in 8.3.0

PR #1016, opened 2023-11-08
Upload progress incorrect in Google Chrome

### LINKED ISSUE #1013: Progress goes straight to 100% when starting a file upload
Had this reported to me via Discord DM.

Suspect there is variance between the HTTP responses, timing etc of different users' upload endpoints, possibly includes browser implementation variance also.

Ref:
#1002
#1005
#1007

```[tasklist]
### Suggested way forward
- [x] Set up a unit test scenario for the internal upload function's request callbacks and how they affect `file.loaded` and `file.progress` (This will likely require some internal refactoring to separate as these are coupled at present)
- [x] Gather data from real upload requests and set up scenario tests with these (example below)
```

Upload telemetry could be gathered by [adding logging to the upload function](https://github.com/adopted-ember-addons/ember-file-upload/blob/master/ember-file-upload/src/system/upload.ts#L90-L121) directly in `node_modules`

image

mkszepp:
Sometimes the browser doesn't has the correct file size `onloadstart`. like described here: https://github.com/adopted-ember-addons/ember-file-upload/pull/1002#issuecomment-1801198715

During debugging i have found out, that this bug is present in Chrome. In Chrome there is often a total wrong `total` size (see [comment](https://github.com/adopted-ember-addons/ember-file-upload/pull/1002#issuecomment-1801264134)) and in Firefox there is sometime a little difference between the total size `onloadstart` and `onprogress`.

We should set `file.size` like in earlier versions also in `onprogress` event and check in `progress` event, if the upload is already complete.

#1013

Progress log on Chrome with this fix (3 files in queue):
```
0
0
1
1
3
3
5
5
6
6
8
8
10
10
11
11
13
13
15
15
16
16
18
18
20
20
21
21
23
23
24
24
24
24
24
24
24
24 // first file uploaded
26
26
27
27
29
29
31
31
33
33
34
34
36
36
38
38
39
39
41
41
43
43
44
44
46
46
48
48
49
49
51
51
53
53
55
55
56
56
58
58
60
60
61
61
63
63
65
65
66
66
68
68
70
70
71
71
73
73
75
75
76
76
78
78
80
80
82
82
83
83
85
85
87
87
88
88
88
88
88
88
88 // 2nd file uploaded
88
89
89
91
91
93
93
95
95
96
96
98
98
100
100
100
100 // all files uploaded
```

Progress log in Firefox after fix (3 files in queue):
```
3
6
9
12
15
19
22
24
24 // first file uploaded
27
31
34
37
40
43
47
50
53
56
59
62
66
69
72
75
78
82
85
86
86 // 2nd file uploaded
89
92
95
98
100 // all files uploaded
```

### DIFF
--- a/ember-file-upload/src/system/upload.ts
+++ b/ember-file-upload/src/system/upload.ts
@@ -92,13 +92,20 @@ export function upload(
     if (!evt.lengthComputable || evt.total === 0) return;
 
     file.loaded = evt.loaded;
-    file.size = evt.total;
+    // It occurs that the evt.total is not always correct.
+    // For this reason we should hold the max file size.
+    // The correct should be returned while progress
+    file.size = Math.max(file.size, evt.total);
     file.progress = (file.loaded / file.size) * 100;
   };
 
   request.onprogress = function (evt) {
     if (!evt) return;
-    if (!evt.lengthComputable || evt.total === 0) return;
+    // We need to check also for isUploadComplete, because the browsers brings sometimes the onprogress after onloadend event
+    if (!evt.lengthComputable || evt.total === 0 || file.isUploadComplete)
+      return;
+
+    file.size = evt.total;
 
     // When the progress is completed there is possible that we get the `Content-Length` response header of the upload endpoint as loaded / total.
     // There is possible that `Content-Length` is lower or higher than the already loaded bytes.
@@ -118,6 +125,7 @@ export function upload(
 
     file.loaded = file.size;
     file.progress = (file.loaded / file.size) * 100;
+    file.isUploadComplete = true;
   };
 
   request.ontimeout = () => {
--- a/ember-file-upload/src/upload-file.ts
+++ b/ember-file-upload/src/upload-file.ts
@@ -87,6 +87,11 @@ export class UploadFile {
    * range of 0 to 100.
    */
   @tracked progress = 0;
+  
+  /**
+   * When upload has finished this property will be set to true
+   */
+  @tracked isUploadComplete = false;
 
   /**
    * The current state that the file is in.
--- a/ember-file-upload/src/upload-file.ts
+++ b/ember-file-upload/src/upload-file.ts
@@ -87,7 +87,7 @@ export class UploadFile {
    * range of 0 to 100.
    */
   @tracked progress = 0;
-  
+
   /**
    * When upload has finished this property will be set to true
    */


RobbieTheWagner:
@mkszepp would it be possible to add some tests for checking progress values? We could then run those tests in both Chrome and Firefox to catch regressions here.

jelhan:
> We could then run those tests in both Chrome and Firefox to catch regressions here.

I'm not sure if we can test the difference at this level. If I understand correctly, both Chrome and Firefox may have a wrong file size in `onloadstart` event. I fear we are mocking that layer in the tests: https://github.com/adopted-ember-addons/ember-file-upload/blob/c3af01e9c4a875fb3c2bf02a02a08b401eaa1576/ember-file-upload/src/mirage/upload-handler.ts#L55-L61

jelhan:
I had a look in specification: https://xhr.spec.whatwg.org/#the-send()-method

If I understand correctly, the `total` of `ProgressEvent` fired at `loadstart` should be either `0` or the actual length of the body. See step 11. But if I understand correctly, we see different behavior of the browsers in reality, correct?

mkszepp:
@jelhan exact, we have specially on Chrome the behavior, that `loadstart` brings a the wrong `total`. I think this problem occurs with large files, small files are not the problems (because there was 1 of 3 files with this bug)...
In Firefox i have got also yesterday with one file a little difference.

RobbieTheWagner:
@mkszepp great, thanks! Please keep me posted on writing tests. I do think we should probably merge this either way, but would love to have regression coverage.

gilest:
> I'm not sure if we can test the difference at this level. If I understand correctly, both Chrome and Firefox may have a wrong file size in onloadstart event. I fear we are mocking that layer in the tests

Yes, that's why I am requesting (in #1013) that we record some live progress data from different browsers and endpoints so that I can re-arrange some internals and then replay those events as test scenarios

gilest:
Pretty sure I've validated with unit tests that this fix is correct – at least for this specific issue (progress straight to 100% in Google Chrome). I'll open a proof PR shortly

gilest:
Thanks again.

Fix released in [8.3.1](https://github.com/adopted-ember-addons/ember-file-upload/releases/tag/v8.3.1)

mkszepp:
Thanks! i have tested in my app and works!

@RobbieTheWagner maybe you can give us also a short feedback, if the bug in you app is also resolved with this fix.

PR #1017, opened 2023-11-10
Replay progress events in unit tests

### LINKED ISSUE #1013: Progress goes straight to 100% when starting a file upload

gilest:
Skips the browser, http, network etc. and lets us test the smallest unit of our internals – an `UploadFile` having its state updated by a series of `ProgressEvent` events.

For each scenario I'm pretty sure we only need to test a single file with a stream of synchronous events.

Covers some recorded events from different browsers and endpoints. We can add more scenarios here if we come across them.

Specifically, the Chrome scenario proves the fix in #1016 – this was previously failing early with 100% progress, which is the same issue reported by users in #1013

I've painstakingly copied values from log screenshots to setup this existing suite. Hopefully contributors can PR full tests in future to make things a bit easier 😅 

Fixes #1013

image

### DIFF
--- a/test-app/tests/unit/system/upload-progress-test.ts
+++ b/test-app/tests/unit/system/upload-progress-test.ts
@@ -1,0 +1,202 @@
+/* eslint-disable qunit/no-conditional-assertions */
+
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+import { FileSource, UploadFile } from 'ember-file-upload';
+import { type TestContext } from '@ember/test-helpers';
+import { png } from 'test-app/tests/utils/file-data';
+import { onloadstart, onprogress, onloadend } from 'ember-file-upload/internal';
+
+interface Handlers {
+  [key: string]: typeof onloadstart;
+}
+
+const handlers: Handlers = {
+  loadstart: onloadstart,
+  progress: onprogress,
+  loadend: onloadend,
+};
+
+type Step = {
+  assertProgressBefore?: number;
+  event: ProgressEvent;
+  assertProgressAfter?: number;
+};
+
+const replayEvents = (
+  assert: Assert,
+  uploadFile: UploadFile,
+  steps: Step[],
+) => {
+  for (const step of steps) {
+    if (typeof step.assertProgressBefore === 'number') {
+      assert.strictEqual(
+        uploadFile.progress,
+        step.assertProgressBefore,
+        `before (index: ${steps.indexOf(step)}), progress is: ${
+          step.assertProgressBefore
+        }% type: ${step.event.type}, total: ${step.event.total}, loaded: ${
+          step.event.loaded
+        }`,
+      );
+    }
+
+    const handler = handlers[step.event.type];
+    if (!handler) throw new Error(`No handler matched ${step.event.type}`);
+
+    handler(uploadFile, step.event);
+
+    if (typeof step.assertProgressAfter === 'number') {
+      assert.strictEqual(
+        uploadFile.progress,
+        step.assertProgressAfter,
+        `after (index: ${steps.indexOf(step)}), progress is: ${
+          step.assertProgressAfter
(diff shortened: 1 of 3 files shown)


RobbieTheWagner:
Thanks for the quick turnaround on this @gilest! Excited to see this come together 😄

RobbieTheWagner:
Thanks for the quick turnaround on this @gilest! Excited to see this come together 😄

test-app/tests/unit/system/upload-progress-test.ts
```diff
@@ -0,0 +1,202 @@
+/* eslint-disable qunit/no-conditional-assertions */
```
gilest, line 1: Not sure if there's a good way to avoid this 🤔 

The data structure and complexity is a bit high for `assert.step` and `assert.verifySteps`.

I'm pretty confident that `replayEvents` is simple enough that assertions are running when we expect them to.

test-app/tests/unit/system/upload-progress-test.ts
```diff
@@ -0,0 +1,202 @@
...
+          total: 618122,
+        }),
+        assertProgressAfter: 5.301218853236093,
+      },
+    ];
+
+    replayEvents(assert, uploadFile, steps);
+  });
+
+  test('Firefox (from issue #1002)', function (this: TestContext, assert) {
```
gilest, line 158: https://user-images.githubusercontent.com/19845290/281305620-5bc45c54-8607-4c3a-88c3-c317c7cc1aee.png

test-app/tests/unit/system/upload-progress-test.ts
```diff
@@ -0,0 +1,202 @@
...
+          total: 390,
+        }),
+        assertProgressAfter: 100,
+      },
+    ];
+
+    replayEvents(assert, uploadFile, steps);
+  });
+
+  test('Chrome (from issue #1002)', function (this: TestContext, assert) {
```
gilest, line 114: https://user-images.githubusercontent.com/19845290/281305327-9240f787-1177-4ca4-b11f-e61f0f732c6a.png

test-app/tests/unit/system/upload-progress-test.ts
```diff
@@ -0,0 +1,202 @@
...
+        }`,
+      );
+    }
+  }
+};
+
+module('Unit | upload progress', function (hooks) {
+  setupTest(hooks);
+
+  test('Original (from issue #1002)', function (this: TestContext, assert) {
```
gilest, line 66: https://user-images.githubusercontent.com/36919/280431274-c69149b9-9918-4feb-9d5d-699a6dc59a3c.png

gilest:
@mkszepp please review ☝🏻

mkszepp:
@gilest looks good, added two comments in code...
We should also add a test in which `loadend` brings `total: 0`, because this would be happen, when sombody returns only status 2xx with no content

gilest:
> We should also add a test in which `loadend` brings `total: 0`, because when sombody returns only status 2xx with no content it occurs

Ah yes – great idea. I have added a test for this.

> @gilest looks good, added two comments in code...

Thanks – not sure if your code comments got saved, I cannot see any...

I'm going to merge and release but feel free to open a tidy-up PR with your suggestions

PR #1024, opened 2023-11-18
Website demo – support HTTP uploads

gilest:
This will be useful for debugging issues and testing new features.

Users being able to interactively explore this addon should help them understand and visualise the less-documented functionality like queue flushing.

### UI

image

### Example HTTP upload

image

### DIFF
--- a/website/app/components/options-form.gjs
+++ b/website/app/components/options-form.gjs
@@ -1,0 +1,120 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { on } from '@ember/modifier';
+import { htmlSafe } from '@ember/template';
+
+export const UPLOAD_TYPES = {
+  simulated: 'Simulated 🧑‍🔬',
+  http: 'HTTP 📡',
+};
+
+// Simulated
+// Values in kilobits per second (kbps)
+const RATES = {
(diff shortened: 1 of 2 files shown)


website/app/components/options-form.gjs
```diff
@@ -0,0 +1,120 @@
...
+const DEFAULT_RATE = RATES['Fast 3G - 0.675 Mbps'];
+
+// HTTP
+const DEFAULT_URL = 'https://api.bytescale.com/v1/files/basic';
+const METHODS = ['GET', 'POST', 'PUT', 'PATCH', 'HEAD'];
+const DEFAULT_METHOD = 'POST';
+const DEFAULT_HEADERS = {
+  Authorization: 'Basic YXBpa2V5OmZyZWU',
+  'Content-Type': 'application/x-www-form-urlencoded',
+};
```
gilest, line 31: While looking for a public "black hole" upload endpoint I found this documented in the MIT npm package `react-uploader` [README](https://www.npmjs.com/package/react-uploader#-file-management-api)

It seems to work for me, but it's a suggestion that user's may want to use. They are using this API on their own behalf 🤷🏻

gilest: Ended up replacing this with a WireMock endpoint from a free account I created

PR #1030, opened 2023-12-01
Move @embroider/macros to peerDeps

RobbieTheWagner:
I am not sure if this actually helps or not, but we noticed a lot of `lodash` being included in our app after installing ember-file-upload and my hypothesis is @embroider/macros should be a peerDep since it is just used in the mirage stuff and that the lodash is coming from it, but I could be totally wrong.

### DIFF
--- a/ember-file-upload/package.json
+++ b/ember-file-upload/package.json
@@ -50,8 +50,7 @@
   },
   "dependencies": {
     "@ember/test-waiters": "^3.0.0",
-    "@embroider/addon-shim": "^1.5.0",
-    "@embroider/macros": "^1.0.0",
+    "@embroider/addon-shim": "^1.8.7",
     "ember-auto-import": "^2.0.0"
   },
   "devDependencies": {
@@ -64,6 +63,7 @@
     "@ember/string": "^3.0.1",
     "@ember/test-helpers": "^3.1.0",
     "@embroider/addon-dev": "^4.0.0",
+    "@embroider/macros": "^1.0.0",
     "@glimmer/component": "^1.0.4",
     "@glimmer/tracking": "^1.0.4",
     "@glint/core": "^1.0.2",
@@ -94,6 +94,7 @@
   },
   "peerDependencies": {
     "@ember/test-helpers": "^2.9.3 || ^3.0.3",
+    "@embroider/macros": "^1.0.0",
     "@glimmer/component": "^1.1.2",
     "@glimmer/tracking": "^1.1.2",
     "ember-cli-mirage": "*",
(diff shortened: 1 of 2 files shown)


gilest:
Hmm I'm not sure about this one... In the readme for that package it says

gilest:
Hmm I'm not sure about this one... In the [readme for that package](https://github.com/embroider-build/embroider/tree/main/packages/macros#setting-configuration-from-an-ember-addon) it explicitly says to include in `dependencies`.

> we noticed a lot of lodash being included in our app after installing ember-file-upload 

If that's the case then this is an issue with `@embroider/maros`, `ember-auto-import` or Embroider itself. We may need to investigate further and raise a bug where relevant.

Are you using the mirage handler in your app? Are you using embroider or classic build?

RobbieTheWagner:
> > we noticed a lot of `lodash` being included in our app after installing ember-file-upload
> 
> Did you noticed it in development or production build? If I recall correctly, Embroider macros work differently for development and production builds.

No build at all, just installing ember-file-upload added a ton of lodash to deps in our pnpm lock.

RobbieTheWagner:
> Hmm I'm not sure about this one... In the [readme for that package](https://github.com/embroider-build/embroider/tree/main/packages/macros#setting-configuration-from-an-ember-addon) it explicitly says to include in `dependencies`.
> 
> Agree that apps which don't use mirage won't need it but it being in the dep tree shouldn't have any negative affects.
> 
> > we noticed a lot of lodash being included in our app after installing ember-file-upload
> 
> If that's the case then this is an issue with `@embroider/maros`, `ember-auto-import` or Embroider itself. We may need to investigate further and raise a bug where relevant.
> 
> Are you using the mirage handler in your app? Are you using embroider or classic build?

We're not using embroider, just classic build.

RobbieTheWagner:
> > No build at all, just installing ember-file-upload added a ton of lodash to deps in our pnpm lock.
> 
> Why is that a problem?

Because they are deps and added to the bundle. If they were devDeps, it would be no problem.

RobbieTheWagner:
> > Because they are deps and added to the bundle.
> 
> Not all code of installed dependencies for into the bundle. Actually only a very small subset does. If everything works as expected, `@embroider/macros` should not push any code into production build. But it is expected to be installed. It is needed at build-time.
> 
> Please check if it affects the production bundle or not. It's only a problem of unexpected code is pushed I to the production build.

It's weird that all the lodash things went from being devDeps to deps when we installed ember-file-upload though. I consider that incorrect if they aren't going into the build.

PR #1048, opened 2023-12-23
Setup `ember-template-imports` in test app

gilest:
Good dogfooding exercise to use the addon in "strict mode".

We may need to drop `3.25` support or split out the tests/test app somehow


test-app/package.json
```diff
@@ -65,6 +66,7 @@
     "ember-resolver": "^11.0.1",
     "ember-source": "~5.4.0",
     "ember-source-channel-url": "^3.0.0",
+    "ember-template-imports": "^3.0.0",
```
gilest, line 69: I _think_ Glint is incompatible with v4 for now

PR #1050, opened 2023-12-23
Drop support for Ember 3.25

gilest:
Raises minimum supported version to 3.28

Purpose is to allow us to use "strict" mode in the test app (#1048 #1049)

Feel this is reasonable as we still support the final release in 3.x series

Thinking I will merge but not bother releasing until we have other changes to release also

Would be good if there were other breaking changes to batch, but I'm not aware of any

### DIFF
--- a/ember-file-upload/README.md
+++ b/ember-file-upload/README.md
@@ -10,15 +10,11 @@ Uploads can be managed through queues and continue in the background, even after
 
 ## Compatibility
 
-* Ember.js 3.25 or above
+* Ember.js 3.28 or above
 * TypeScript 5.0 or above
 * ember-auto-import 2.0 or above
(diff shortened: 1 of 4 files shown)


ember-file-upload/README.md
```diff
@@ -10,16 +10,12 @@ Uploads can be managed through queues and continue in the background, even after
 
 ## Compatibility
 
-* Ember.js 3.25 or above
+* Ember.js 3.28 or above
 * TypeScript 5.0 or above
 * ember-auto-import 2.0 or above
-* Ember CLI v2.13 or above
-* Node.js v16 or above
```
gilest, line 17: I don't have a good reference for this but since this is a v2 addon Ember CLI and Node versions are not relevant to compatibility

PR #1066, opened 2024-01-31
feat: allow dropping a folder into the dropzone

gvdp:
Recreated https://github.com/adopted-ember-addons/ember-file-upload/pull/834 after updating master with all the latest changes.

As for the question about how in-demand this is: we've needed it for one of our apps and it's been in production for over a year now. I made it so there is an opt-in boolean which needs to be enabled, so by default nothing should change for exisiting users.

### DIFF
--- a/ember-file-upload/src/components/file-dropzone.ts
+++ b/ember-file-upload/src/components/file-dropzone.ts
@@ -65,6 +65,10 @@ export default class FileDropzoneComponent extends Component<FileDropzoneSignatu
     return this.args.multiple ?? true;
   }
 
+  get allowFolderDrop() {
+    return this.args.allowFolderDrop ?? false;
+  }
+
   get files(): File[] {
     const files = this.dataTransferWrapper?.files ?? [];
     if (this.multiple) return files;
@@ -134,7 +138,7 @@ export default class FileDropzoneComponent extends Component<FileDropzoneSignatu
   }
 
   @action
-  didDrop(event: FileUploadDragEvent) {
+  async didDrop(event: FileUploadDragEvent) {
     if (this.dataTransferWrapper) {
       this.dataTransferWrapper.dataTransfer = event.dataTransfer;
     }
@@ -219,9 +223,15 @@ export default class FileDropzoneComponent extends Component<FileDropzoneSignatu
     // }
 
     if (this.dataTransferWrapper) {
-      const addedFiles = this.addFiles(this.files);
-      this.args.onDrop?.(addedFiles, this.dataTransferWrapper);
+      let files;
+      if (this.allowFolderDrop) {
+        files = await this.dataTransferWrapper.getFilesAndDirectories();
+      } else {
+        files = this.files;
+      }
 
+      const addedFiles = this.addFiles(files);
+      this.args.onDrop?.(addedFiles, this.dataTransferWrapper);
       this.active = false;
       this.dataTransferWrapper = undefined;
     }
--- a/ember-file-upload/src/interfaces.ts
+++ b/ember-file-upload/src/interfaces.ts
@@ -120,6 +120,14 @@ export interface FileDropzoneSignature {
     allowUploadsFromWebsites?: boolean;
 
     /**
+     *
+     * Wether users can drop folders into the dropzone
+     *
+     * @defaulValue false
+     * */
+    allowFolderDrop?: boolean;
+
+    /**
      * This is the type of cursor that should
      * be shown when a drag event happens.
      *
--- a/ember-file-upload/src/system/data-transfer-wrapper.ts
+++ b/ember-file-upload/src/system/data-transfer-wrapper.ts
@@ -43,6 +43,69 @@ export default class DataTransferWrapper {
     return this.files.length ? this.files : this.items;
   }
 
+  async getFilesAndDirectories() {
+    const files: File[] = [];
+
+    const readEntry = async (entry: FileSystemEntry): Promise<File> => {
+      return new Promise((resolve, reject) => {
+        if (entry.isFile) {
+          (entry as FileSystemFileEntry).file((fileEntry: File) => {
+            resolve(fileEntry);
+          });
+        } else {
+          reject('Directory contains nested directories');
+        }
+      });
+    };
+
+    const readAllFilesInDirectory = (item: DataTransferItem): Promise<File[]> =>
+      new Promise((resolve) => {
+        (item.webkitGetAsEntry() as FileSystemDirectoryEntry)
+          ?.createReader()
+          ?.readEntries(async (entries: FileSystemEntry[]) => {
+            const readFiles: File[] = await Promise.all(
+              entries.map(readEntry),
+            ).catch((err) => {
+              throw err;
+            });
+            resolve(readFiles.filter((file) => file !== undefined) as File[]);
+          });
+      });
+
+    const readDataTransferItem = async (
+      item: DataTransferItem,
+    ): Promise<File[]> => {
+      if (item.webkitGetAsEntry()?.isDirectory) {
+        const directoryFile = item.getAsFile() as File;
+        const filesInDirectory: File[] = await readAllFilesInDirectory(item);
+        return [directoryFile, ...filesInDirectory];
+      } else {
+        const fileItem = item.getAsFile() as File;
+        return [fileItem];
+      }
+    };
+
+    if (this.dataTransfer?.items) {
+      const allFilesInDataTransferItems: File[][] = await Promise.all(
+        Array.from(this.dataTransfer?.items).map(readDataTransferItem),
+      );
+
+      const flattenedFileArray: File[] = allFilesInDataTransferItems.reduce(
+        (flattenedList, fileList) => {
+          return [...flattenedList, ...fileList];
+        },
+        [],
+      );
+
+      files.push(...flattenedFileArray);
+    } else {
+      const droppedFiles: File[] = Array.from(this.dataTransfer?.files ?? []);
+      files.push(...droppedFiles);
+    }
+
+    return files;
+  }
+
   get files() {
     return Array.from(this.dataTransfer?.files ?? []);
   }
--- a/ember-file-upload/src/test-support.ts
+++ b/ember-file-upload/src/test-support.ts
@@ -91,6 +91,60 @@ export async function dragAndDrop(
   return triggerEvent(dropzone, 'drop', { dataTransfer });
 }
 
+interface FileSystemEntryStub {
+  isFile: boolean;
+  file: (callback: (file: File | Blob) => void) => void;
+}
+
+export async function dragAndDropDirectory(
+  selector: string,
+  folderName: string,
+  filesInDirectory: (File | Blob)[],
+  ...singleFiles: (File | Blob)[]
+) {
+  const dropzone = find(selector);
+  assert(`Selector '${dropzone}' could not be found.`, dropzone);
+  assert(
+    'All files must be instances of File/Blob type',
+    filesInDirectory.every((file) => file instanceof Blob),
+  );
+
+  const folderItem = {
+    webkitGetAsEntry: () => ({
+      isDirectory: true,
+      createReader: () => ({
+        readEntries: (callback: (entries: FileSystemEntryStub[]) => void) => {
+          const entryFiles = filesInDirectory.map((file) => {
+            return {
+              isFile: true,
+              file: (callback: (file: File | Blob) => void) => {
+                callback(file);
+              },
+            };
+          });
+          callback(entryFiles);
+        },
+      }),
+    }),
+    getAsFile: () => new File([], folderName, { type: '' }),
+  };
+
+  const singleFileItem = (singleFile: File | Blob) => ({
+    webkitGetAsEntry: () => ({
+      isDirectory: false,
+    }),
+    getAsFile: () => singleFile,
+  });
+
+  const dataTransfer = {
+    items: [folderItem, ...singleFiles.map(singleFileItem)],
+  };
+
+  await triggerEvent(dropzone, 'dragenter', { dataTransfer });
+  await triggerEvent(dropzone, 'dragover', { dataTransfer });
+  return triggerEvent(dropzone, 'drop', { dataTransfer });
+}
+
 /**
   Triggers a `dragenter` event on a `FileDropzone` with `files`.
 
--- a/test-app/tests/integration/components/file-dropzone-test.gts
+++ b/test-app/tests/integration/components/file-dropzone-test.gts
@@ -3,6 +3,7 @@ import { setupRenderingTest } from 'ember-qunit';
 import { render, triggerEvent, type TestContext } from '@ember/test-helpers';
 import {
   dragAndDrop,
+  dragAndDropDirectory,
   dragEnter,
   dragLeave,
 } from 'ember-file-upload/test-support';
@@ -177,6 +178,28 @@ module('Integration | Component | FileDropzone', function (hooks) {
     );
 
     assert.verifySteps(['dingus.txt']);
+  });
+
+  test('allowFolderDrop=true allows dropping a directory and reads out the content', async function (this: LocalTestContext, assert) {
+    const queue = this.queue;
+    const onDrop = (files: UploadFile[]) => {
+      files.forEach((file) => assert.step(file.name));
+    };
+
+    await render(<template>
+        <FileDropzone
+          class="test-dropzone"
+          @queue={{queue}}
+          @allowFolderDrop={{true}}
+          @onDrop={{onDrop}} />
+      </template>);
+
+    await dragAndDropDirectory('.test-dropzone', 'directory-name', [
+      new File([], 'dingus.txt'),
+      new File([], 'dingus.png'),
+    ]);
+
+    assert.verifySteps(['directory-name', 'dingus.txt', 'dingus.png']);
   });
 
   // Check for regression of: https://github.com/adopted-ember-addons/ember-file-upload/issues/446
--- a/test-app/app/components/demo-upload.gts
+++ b/test-app/app/components/demo-upload.gts
@@ -71,7 +71,7 @@ export default class DemoUpload extends Component<DemoUploadSignature> {
       as |queue|
     }}
       <div class='docs-my-8 text-center'>
-        <FileDropzone @queue={{queue}} class='demo-dropzone' as |dropzone|>
+        <FileDropzone @queue={{queue}} @allowFolderDrop={{true}} class='demo-dropzone' as |dropzone|>
           <div
             class='dropzone-upload-area upload {{if dropzone.active "active"}}'
           >
--- a/test-app/tests/unit/system/data-transfer-wrapper-test.js
+++ b/test-app/tests/unit/system/data-transfer-wrapper-test.js
@@ -66,4 +66,37 @@ module('Unit | DataTransferWrapper', function (hooks) {
     };
     assert.strictEqual(this.subject.filesOrItems.length, 2);
   });
+
+  test('directory dropped', async function (assert) {
+    const transfer = {
+      items: [
+        folderItem('directory-name', [
+          new File([], 'fileName.txt'),
+          new File([], 'otherFileName.txt'),
+        ]),
+      ],
+    };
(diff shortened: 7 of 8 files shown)


ember-file-upload/src/components/file-dropzone.ts
```diff
@@ -219,9 +223,15 @@ export default class FileDropzoneComponent extends Component<FileDropzoneSignatu
     // }
 
     if (this.dataTransferWrapper) {
-      const addedFiles = this.addFiles(this.files);
-      this.args.onDrop?.(addedFiles, this.dataTransferWrapper);
+      let files;
+      if (this.allowFolderDrop) {
```
gilest, line 227: We can avoid defining a getter for this property because `undefined` is falsy anyway

```suggestion
      if (this.args.allowFolderDrop) {
```

ember-file-upload/src/components/file-dropzone.ts
```diff
@@ -219,9 +223,15 @@ export default class FileDropzoneComponent extends Component<FileDropzoneSignatu
...
+        files = await this.dataTransferWrapper.getFilesAndDirectories();
+      } else {
+        files = this.files;
+      }
```
gilest, line 231: We can avoid defining a getter for this property because `undefined` is falsy anyway

```suggestion
      const files = this.args.allowFolderDrop ?
        await this.dataTransferWrapper.getFilesAndDirectories() :
        this.files;
```

ember-file-upload/src/components/file-dropzone.ts
```diff
@@ -65,6 +65,10 @@ export default class FileDropzoneComponent extends Component<FileDropzoneSignatu
     return this.args.multiple ?? true;
   }
 
+  get allowFolderDrop() {
+    return this.args.allowFolderDrop ?? false;
+  }
+
```
gilest, line 71: We can avoid defining a getter for this property because `undefined` is falsy anyway

```suggestion
```

ember-file-upload/src/interfaces.ts
```diff
@@ -119,6 +119,14 @@ export interface FileDropzoneSignature {
      * */
     allowUploadsFromWebsites?: boolean;
 
+    /**
+     *
+     * Wether users can drop folders into the dropzone
```
gilest, line 124:
```suggestion
     * Whether users can drop folders into the dropzone
```

ember-file-upload/src/test-support.ts
```diff
@@ -91,6 +91,60 @@ export async function dragAndDrop(
...
+  isFile: boolean;
+  file: (callback: (file: File | Blob) => void) => void;
+}
+
+export async function dragAndDropDirectory(
+  selector: string,
+  folderName: string,
+  filesInDirectory: (File | Blob)[],
+  ...singleFiles: (File | Blob)[]
+) {
```
gilest, line 104: I wonder if we should use an options object here. From the call site the two types of files (in directory and single) are not very clear.

test-app/tests/integration/components/file-dropzone-test.gts
```diff
@@ -179,6 +180,28 @@ module('Integration | Component | FileDropzone', function (hooks) {
...
+          @allowFolderDrop={{true}}
+          @onDrop={{onDrop}} />
+      </template>);
+
+    await dragAndDropDirectory('.test-dropzone', 'directory-name', [
+      new File([], 'dingus.txt'),
+      new File([], 'dingus.png'),
+    ]);
+
+    assert.verifySteps(['directory-name', 'dingus.txt', 'dingus.png']);
```
gilest, line 202: Not sure that a directory should be included in the files array from `onDrop`. It's not really an `UploadFile` that is able to be uploaded

We might need to add another property to these callbacks which _only_ has directories (maybe as strings of the directory name)

ember-file-upload/src/system/data-transfer-wrapper.ts
```diff
@@ -43,6 +43,69 @@ export default class DataTransferWrapper {
...
+          });
+        } else {
+          reject('Directory contains nested directories');
+        }
+      });
+    };
+
+    const readAllFilesInDirectory = (item: DataTransferItem): Promise<File[]> =>
+      new Promise((resolve) => {
+        (item.webkitGetAsEntry() as FileSystemDirectoryEntry)
```
gilest, line 63: The [advice from MDN](https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/webkitGetAsEntry) about this method is:

> Note: This function is implemented as webkitGetAsEntry() in non-WebKit browsers including Firefox at this time; it may be renamed to getAsEntry() in the future, so you should code defensively, looking for both.

I'd suggest that we wrap this function and attempt to call `getAsEntry` and then `webkitGetAsEntry`. Hopefully that will be supported in future without us needing to make a change

gilest (commented):
Hey @gvdp. Thank you for reopening this – appreciate your contributions as always ❤️ 

> so by default nothing should change for exisiting users.

That's true.. But adding a feature will increase the bundle size for all addon users. When we add a feature we are committing as maintainers to support it for a long time, too

For testing purposes I have pushed this feature branch to `docs/folder-drop` which deploys a public docsite here:
https://docs-folder-drop.ember-file-upload.pages.dev/demo

I'll keep that updated so that we can test across different browsers and devices.

I did some light testing in Safari, Firefox, Chrome and Edge on Mac OS and Windows and found a few issues:

- [ ] Dropping a directory of directories throws and leaves the dropzone in a broken state (`dropzone.active` stays `true` afterwards)
  - Interested in your suggestion of how we might handle this
    - recursively open subdirectories and add their files, too
    - only add files one-level deep but do not throw an error when nested folders are present
- [x] The directory itself is being reported as an `UploadFile` in callbacks
  - this might be useful programatically, it's a significant API change and something that's not modelled as part of this addon. For example there's no way to "upload" the directory from the queue
  - maybe we can get notified about directory details at the dropzone-event level, but I'm not sure we want those items being added to the queue
- [x] In some browsers (Firefox on Mac OS) `.DS_Store` files are being added. Not sure there's ever a reason this would be desired. I guess we should consider how to handle dot files in general, too

In terms of the implementation, it would be great to reduce the usage of the TypeScript `as` keyword. That would give us more confidence about the robustness of this feature

Once we've resolved these issues, we'll also need to add documentation of this feature
- [ ] API reference -> FileDropzone
- [ ] Documentation -> Testing
- [ ] (New page) Documentation -> Add directories via Drag and Drop

ember-file-upload/src/components/file-dropzone.ts
gvdp: Implemented, removed the getter.

ember-file-upload/src/components/file-dropzone.ts
gvdp: Removed

ember-file-upload/src/interfaces.ts
gvdp: Good catch, corrected.

ember-file-upload/src/test-support.ts
gvdp: Fair call, it's not super clear indeed. Would you make 1 `opts` object? Or keep the `selector` and/or `folderName` and make an object for the rest?

I'm doubting if the `folderName` even needs to be here, but that'll probably be relevant for the bigger remark about the directory itself being reported as `UploadFile`

ember-file-upload/src/system/data-transfer-wrapper.ts
gvdp: Wrapped it in a local `getEntry` function. Typescript does complain because `getAsEntry` is not (yet) in the lib definition. Tried to circumvent this by extending the interface to `FutureProofDataTransferItem` , not sure if that's the cleanest solution.

gvdp:
Hi @gilest , thanks for the comments, appreciate the effort. I fixed some of the smaller code comments already. Not sure how you usually commit those? I made a few small extra commits now but they can be easily squashed.

- The directory of directories is indeed something which does not work. In our own application this is not necessary and we give the end user an explicit notice of this. I don't think including everything recursively is the way to go, maybe we should just throw or show a specific, relevant error to the user?
- The directory itself being reported as `UploadFile` is indeed not ideal, I seem to remember we did this for a reason but it's already some time ago and I don't know what it was :-) . As a `File` object it's not possible to see the difference between a file and a directory, but there is the `isDirectory` prop which we do already use. Maybe we should just not include the `directoryFile` in the array being sent [here](https://github.com/adopted-ember-addons/ember-file-upload/pull/1066/files#diff-f34113bbd5fdfb4955eb5b79ed556bf12ac8d0b8a9f0d0bccb536c87294b9bebR97)? I think this would not break our current implementation. That would probably also mean the `directoryName` prop in the test helper is no longer relevant.
- The `.DS_store` files are also annoying. In our implementation we catch and filter those out in the consuming application. Not sure we should just ignore all dotfiles, sometimes it can be intentional that these are uploaded. Maybe we should just include a warning in the documentation that it is the responsibility of the consuming application to handle all hidden files?

I'll of course also add documentation and API reference stuff once we've resolved these open issues. Thanks again for the review!

ember-file-upload/src/system/data-transfer-wrapper.ts
```diff
@@ -43,6 +47,81 @@ export default class DataTransferWrapper {
...
+      new Promise((resolve) => {
+        getEntry(item)
+          ?.createReader()
+          ?.readEntries(async (entries: FileSystemEntry[]) => {
+            const readFiles: File[] = await Promise.all(
+              entries.map(readEntry),
+            ).catch((err) => {
+              throw err;
+            });
+            resolve(readFiles.filter((file) => file !== undefined) as File[]);
```
gilest, line 87: Safer to filter out missing elements with a [truthiness](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) check

```suggestion
            resolve(readFiles.filter(Boolean) as File[]);
```

ember-file-upload/src/system/data-transfer-wrapper.ts
```diff
@@ -43,6 +47,81 @@ export default class DataTransferWrapper {
...
+        [],
+      );
+
+      files.push(...flattenedFileArray);
+    } else {
+      const droppedFiles: File[] = Array.from(this.dataTransfer?.files ?? []);
+      files.push(...droppedFiles);
+    }
+
+    return files;
```
gilest, line 122: Not sure we need to define a files variable and mutate it. Each of these branches could just be an outright return

ember-file-upload/src/system/data-transfer-wrapper.ts
```diff
@@ -43,6 +47,81 @@ export default class DataTransferWrapper {
     return this.files.length ? this.files : this.items;
   }
 
+  async getFilesAndDirectories() {
```
gilest, line 50: [Perf] All of the child functions defined within `getFilesAndDirectories` do not rely on class state or local variable closure.

Because of this, they only need to be defined once in memory, rather than every time this methodn is called.

Let's move them to module scope (you can put them in the top of this file).

I think extracting them also helps with readability of this feature.

ember-file-upload/src/system/data-transfer-wrapper.ts
```diff
@@ -43,6 +47,81 @@ export default class DataTransferWrapper {
...
+    const files: File[] = [];
+
+    const readEntry = async (entry: FileSystemEntry): Promise<File> => {
```
gilest, line 53: This function returns a promise and does not use the `await` keyword.

```suggestion
    const readEntry = (entry: FileSystemEntry): Promise<File> => {
```

ember-file-upload/src/test-support.ts
```diff
@@ -91,6 +91,60 @@ export async function dragAndDrop(
   return triggerEvent(dropzone, 'drop', { dataTransfer });
 }
 
+interface FileSystemEntryStub {
+  isFile: boolean;
+  file: (callback: (file: File | Blob) => void) => void;
+}
+
+export async function dragAndDropDirectory(
+  selector: string,
```
gilest, line 100: Let's follow the same convention as the other test helpers and allow this to be a selector or an element.

```suggestion
    target: string | HTMLElement,
```

gilest (commented):
As always, thanks for your continued work on this.

Sorry this review took so long – it's hard to find a chunk of time to dedicate to a complex feature like this.

For testing purposes I have pushed this feature branch to docs/folder-drop which deploys a public docsite here:
https://docs-folder-drop.ember-file-upload.pages.dev/demo

Have made some general suggestions, especially around improving resilience when a file fails to read.

Will respond to remaining design questions in a separate comment.

gilest:
## Separate comment with "design feedback"

> I fixed some of the smaller code comments already. Not sure how you usually commit those? I made a few small extra commits now but they can be easily squashed.

Extra commits is no problem. The whole feature will be squash-merged anyway 🙂 

> * The directory of directories is indeed something which does not work. In our own application this is not necessary and we give the end user an explicit notice of this. I don't think including everything recursively is the way to go, maybe we should just throw or show a specific, relevant error to the user?

My suggestion: let's add the simplest implementation for now, which we can extend later (ideally without too much breaking change).

Given that, I think it would be ok include _only_ files from the top-level of the dropped directory, and ignore subdirectories. 

This is more flexible than any nested directories causing an error. For example, this avoids cause users to have to re-arrange their local directories just to upload a group of files which would be very frustrating.

> * The directory itself being reported as `UploadFile` is indeed not ideal, I seem to remember we did this for a reason but it's already some time ago and I don't know what it was :-) . As a `File` object it's not possible to see the difference between a file and a directory, but there is the `isDirectory` prop which we do already use. Maybe we should just not include the `directoryFile` in the array being sent [here](https://github.com/adopted-ember-addons/ember-file-upload/pull/1066/files#diff-f34113bbd5fdfb4955eb5b79ed556bf12ac8d0b8a9f0d0bccb536c87294b9bebR97)? I think this would not break our current implementation. That would probably also mean the `directoryName` prop in the test helper is no longer relevant.

Given the "simplest implementation" suggestion I would agree with this and prefer that we not report directories for now. We can model this later as an enhancement (along with recursion).

> * The `.DS_store` files are also annoying. In our implementation we catch and filter those out in the consuming application. Not sure we should just ignore all dotfiles, sometimes it can be intentional that these are uploaded. Maybe we should just include a warning in the documentation that it is the responsibility of the consuming application to handle all hidden files?

Agree with your suggestion – worth documenting. We shouldn't make this decision for users as we can't anticipate their use-cases.

gvdp:
> As always, thanks for your continued work on this.
> 
> Sorry this review took so long – it's hard to find a chunk of time to dedicate to a complex feature like this.
> 
> For testing purposes I have pushed this feature branch to docs/folder-drop which deploys a public docsite here: https://docs-folder-drop.ember-file-upload.pages.dev/demo
> 
> Have made some general suggestions, especially around improving resilience when a file fails to read.
> 
> Please don't be discouraged by my feedback 😅 This is library code so it's better that we consider failure modes and use-cases now than make breaking changes later
> 
> Will respond to remaining design questions in a separate comment.

No problem at all, I appreciate the feedback and I agree it's better to be thorough now. You're the one who probably has to maintain all this later :-).

ember-file-upload/src/test-support.ts
gvdp: Updated

ember-file-upload/src/system/data-transfer-wrapper.ts
gvdp: Good catch. Removed the async keyword. 
There is an eslint rule which can alert these cases, maybe you can add this to you configuration:
https://eslint.org/docs/latest/rules/require-await

ember-file-upload/src/system/data-transfer-wrapper.ts
gvdp: Moved all the functions to the top.

ember-file-upload/src/system/data-transfer-wrapper.ts
gvdp: Made it immediate returns

ember-file-upload/src/system/data-transfer-wrapper.ts
gvdp: Didn't know you could use the `Boolean` constructor like this. Updated.

ember-file-upload/src/system/data-transfer-wrapper.ts
```diff
@@ -43,6 +47,81 @@ export default class DataTransferWrapper {
...
+      }
+
+      return item.webkitGetAsEntry() as FileSystemDirectoryEntry;
+    };
+
+    const readAllFilesInDirectory = (item: DataTransferItem): Promise<File[]> =>
+      new Promise((resolve) => {
+        getEntry(item)
+          ?.createReader()
+          ?.readEntries(async (entries: FileSystemEntry[]) => {
```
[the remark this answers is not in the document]
gvdp: Added a Promise rejection when getEntry is falsy. Not sure it should maybe have a clearer message.

ember-file-upload/src/system/data-transfer-wrapper.ts
```diff
@@ -2,6 +2,10 @@ import type { FileUploadDragEvent } from '../interfaces.ts';
 
 const getDataSupport = {};
 
+interface FutureProofDataTransferItem extends DataTransferItem {
+  getAsEntry?: () => FileSystemDirectoryEntry;
+}
```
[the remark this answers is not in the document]
gvdp: Added the `null` case

gvdp:
> ## Separate comment with "design feedback"
> > I fixed some of the smaller code comments already. Not sure how you usually commit those? I made a few small extra commits now but they can be easily squashed.
> 
> Extra commits is no problem. The whole feature will be squash-merged anyway 🙂
> 
> > * The directory of directories is indeed something which does not work. In our own application this is not necessary and we give the end user an explicit notice of this. I don't think including everything recursively is the way to go, maybe we should just throw or show a specific, relevant error to the user?
> 
> My suggestion: let's add the simplest implementation for now, which we can extend later (ideally without too much breaking change).
> 
> Given that, I think it would be ok include _only_ files from the top-level of the dropped directory, and ignore subdirectories.
> 
> This is more flexible than any nested directories causing an error. For example, this avoids cause users to have to re-arrange their local directories just to upload a group of files which would be very frustrating.
> 
> > * The directory itself being reported as `UploadFile` is indeed not ideal, I seem to remember we did this for a reason but it's already some time ago and I don't know what it was :-) . As a `File` object it's not possible to see the difference between a file and a directory, but there is the `isDirectory` prop which we do already use. Maybe we should just not include the `directoryFile` in the array being sent [here](https://github.com/adopted-ember-addons/ember-file-upload/pull/1066/files#diff-f34113bbd5fdfb4955eb5b79ed556bf12ac8d0b8a9f0d0bccb536c87294b9bebR97)? I think this would not break our current implementation. That would probably also mean the `directoryName` prop in the test helper is no longer relevant.
> 
> Given the "simplest implementation" suggestion I would agree with this and prefer that we not report directories for now. If needed, we can model this later as an enhancement (along with recursion).
> 
> > * The `.DS_store` files are also annoying. In our implementation we catch and filter those out in the consuming application. Not sure we should just ignore all dotfiles, sometimes it can be intentional that these are uploaded. Maybe we should just include a warning in the documentation that it is the responsibility of the consuming application to handle all hidden files?
> 
> Agree with your suggestion – worth documenting. We shouldn't make this decision for users as we can't anticipate their use-cases.

I removed the Promise rejection when there is a subdirectory so the user doesn't get an unnecessary error. I am wondering if we should log a warning or something in this case or just silently ignore this case? Or will just documenting this behavior be enough? I already added a small code comment as well.

Also removed the reporting of the dropped directory itself, now only the files included will be reported. I think I remember now that we used this to keep track of the name of the dropped folder, but that's not super important and we can solve on our end.
Everything still seems to work for me on the demo page :-)

Unrelated comment: I noticed while working on this some of the scripts in the root package.json don't quite work. When you run `pnpm run test` there is a `prepare` script being executed which doesn't exist, and when running `pnpm test:watch:test-app` the `--server` option also doesn't exist.

gilest:
Hi @gvdp . Thank you for your comment and contribution. I had completely missed this notification in a sea of others. Apologies. I will try to review soon!

PR #1076, opened 2024-09-09
Allow `@ember/test-helpers` v4 as peerDependency

mkszepp:
`@ember/test-helpers` was recently updated to a v2 addon. Atm we are getting in our app peerDependency warnings of this addon, because it doesn't allow `@ember/test-helpers` v4.


gilest (approved):
Very nice. Thank you

PR #1084, opened 2024-09-25
fix: @ember/test-helpers `setResolver` or `setApplication`

### LINKED ISSUE #1083: Problem with @ember/test-helpers>4.0.0
## How to reproduce
1. Clone this repo at SHA: `5bc0e4a374fb6c039a62a0c98b82f5448a175d10`
2. `pnpm install`
3. `cd test-app`
4. `ember serve`
5. GOTO: http://localhost:4200/tests
6. Observe that many tests fail with:
```
Promise rejected before "can access files": You must set up the ember-test-helpers environment with either `setResolver` or `setApplication` before running any tests.
```

MichalBryxi:
- Without this patch many tests fail for me with:

```
Promise rejected before "can access files": 
You must set up the ember-test-helpers environment 
with either `setResolver` or `setApplication` 
before running any tests.
```

- Fixes #1083
- Might need https://github.com/adopted-ember-addons/ember-file-upload/pull/1082 to be applied before this fix is useable

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,11 @@ jobs:
           node-version: '${{ env.NODE_VERSION }}'
           args: --no-lockfile
 
+      - name: Install Dependencies
+        run: pnpm install --frozen-lockfile
+      - run: pnpm build
+      - run: pnpm i -f # re-sync injected deps
+
       - name: Test
         run: pnpm --filter test-app test:ember --launch chrome
 
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,2 @@
 resolution-mode=highest
+shared-workspace-lockfile=false
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -160,8 +160,8 @@ importers:
         specifier: ^4.0.0
         version: 4.0.0
       '@ember/test-helpers':
-        specifier: ^3.3.1
-        version: 3.3.1(@babel/core@7.25.2)(@glint/template@1.3.0)(ember-source@5.11.0(@glimmer/component@1.1.2(@babel/core@7.25.2))(@glint/template@1.3.0)(rsvp@4.8.5)(webpack@5.94.0))(webpack@5.94.0)
+        specifier: ^4.0.4
+        version: 4.0.4(@babel/core@7.25.2)(@glint/template@1.3.0)(ember-source@5.11.0(@glimmer/component@1.1.2(@babel/core@7.25.2))(@glint/template@1.3.0)(rsvp@4.8.5)(webpack@5.94.0))
       '@embroider/test-setup':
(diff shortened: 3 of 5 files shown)


MichalBryxi:
The weird thing is that `setApplication` is being called: https://github.com/adopted-ember-addons/ember-file-upload/blob/5bc0e4a374fb6c039a62a0c98b82f5448a175d10/test-app/tests/test-helper.js#L4

mkszepp:
I thinks this is not the correct fix...

With this change we don't test anymore test-helper v4, because it was changes in test-app.

The error occures only inside monorepos (it has no side effect in consumer apps)...

Possible fixes without downgrading in test-app:
- add `pnpm i -f` in ci.yml after build like we did in ember power select see https://github.com/cibernox/ember-power-select/blob/master/.github/workflows/ci.yml#L87
- split lock file by adding in `.npmrc` this `shared-workspace-lockfile=false`
- add test-helper in ember-try into `overrides` instead `devDependencies`

MichalBryxi:
Ok, I actually have no idea what is happening here nor why. Just vague sense that all these steps make it tested via ember-try in correct way against correct versions.

I did not do the "we should add in ember-try a test for test-helper v3" as I could not infer what action should be taken there.

@mkszepp I just pushed the changes in a way I understood from your description. It-s highly likely it's completely wrong so would be really glad if you'd help me out here.

mkszepp:
@MichalBryxi sry for my delay... i have tested locally to reproduce the issue... It looks looke the pnpm-lock.yaml is not correct and you are running for this reason into this issue...

The only thing which you need to to is:
1. revert all changes which were done in PR (or start from master branch)
2. remove all `node_modules` folders
3. remove `pnpm-lock.yaml`
4. run `pnpm install`

After that you can run again your tests and everything should pass...

gilest:
Thanks for the PR @MichalBryxi  and thanks for your analysis @mkszepp 

I will merge the lockfile regen PR. Hopefully this resolves the other local dev issues

Any idea how the lockfile got out of date? Perhaps I merged dependabot PRs too quickly...

MichalBryxi:
Apologies, the PR #1082 only supported this issue, not fixed it. I am used to create PRs for individual fixes.

What I'm seeing on current `master` is still the same issue:

- https://gist.github.com/MichalBryxi/29639e65b823b968d42b0f63ac1cace3

Compare this to an improved situation on _this branch_:

- https://gist.github.com/MichalBryxi/d9f62075d6e6e9af09b30a41c625d11d

mkszepp:
@gilest maybe you can merge my PR, when @MichalBryxi gives a positive feedback

MichalBryxi:
Yay! 👍🏾  I can confirm that #1089 fixes this issue. Thank you!

gilest:
Have done so. Thank you both for reporting, investigating and testing

Issue #1085, opened 2024-09-25
Service cannot be used before helper


MichalBryxi:
## The problem

Given code:

```gts
import Component from '@glimmer/component';
import fileQueue from 'ember-file-upload/helpers/file-queue';

export default class Foo extends Component {
      @service fileQueue;

      <template>
       {{! Block A }}
        <ul>
          {{#each this.fileQueue.files as |file|}}
            <li>
              {{file.name}}
            </li>
          {{/each}}
        </ul>
        
        {{! Block B }}
        {{#let (fileQueue) as |queue|}}
          <label>
            <input type='file' {{queue.selectFile}} hidden />
            Select File
          </label>
        {{/let}}
      </template>
}
```

The app will fail with:

```
ember.js:626 Uncaught (in promise) Error: Assertion Failed: You attempted to update `_value` on `TrackedStorageImpl`, but it had already been used previously in the same computation.  Attempting to update a value after using it in a computation can cause logical errors, infinite revalidation bugs, and performance issues, and is not supported.

`_value` was first used:

- While rendering:
  -top-level
    application
      index
        (unknown template-only component)
          (result of a `unknown` helper)
            (result of a `unknown` helper)
              (result of a `unknown` helper)
                false.files
```

- Curiously enough, if the `Block A` and `Block B` are swapped, the app works just fine

## Versions
-    "ember-file-upload": "^9.1.0",
```sh
ember --version
ember-cli: 5.4.1
node: 20.17.0
os: darwin x64

pnpm --version
9.9.0
```

gilest:
Reopening because the merged fix is only a partial fix

PR #1086, opened 2024-09-25
chore: Service cannot be used before helper failing test

MichalBryxi:
- When `fileQueue` **service** is used _before_ `(fileQueue)` **helper**, then app fails with:

```
ember.js:626 Uncaught (in promise) Error: Assertion Failed: You attempted to update `_value` on `TrackedStorageImpl`, but it had already been used previously in the same computation.  Attempting to update a value after using it in a computation can cause logical errors, infinite revalidation bugs, and performance issues, and is not supported.

`_value` was first used:

- While rendering:
  -top-level
    application
      index
        (unknown template-only component)
          (result of a `unknown` helper)
            (result of a `unknown` helper)
              (result of a `unknown` helper)
                false.files
```
- This PR is just to document the issue described in #1085

Note that for me to run the test suite locally I had to first apply following patches:
- https://github.com/adopted-ember-addons/ember-file-upload/pull/1084
- https://github.com/adopted-ember-addons/ember-file-upload/pull/1082

### DIFF
--- a/test-app/tests/integration/helpers/file-queue-helper-test.gts
+++ b/test-app/tests/integration/helpers/file-queue-helper-test.gts
@@ -5,7 +5,7 @@ import { render, click, settled, waitFor } from '@ember/test-helpers';
 import { DEFAULT_QUEUE, FileState } from 'ember-file-upload';
 import type { UploadFile } from 'ember-file-upload';
 import { selectFiles } from 'ember-file-upload/test-support';
-import { uploadHandler } from 'ember-file-upload';
+import { type FileQueueService, uploadHandler } from 'ember-file-upload';
 import { later } from '@ember/runloop';
 import fileQueue from 'ember-file-upload/helpers/file-queue';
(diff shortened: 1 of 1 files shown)


test-app/tests/integration/helpers/file-queue-helper-test.gts
```diff
@@ -20,6 +20,31 @@ module('Integration | Helper | file-queue', function (hooks) {
   setupRenderingTest(hooks);
   setupMirage(hooks);
 
+  test('service can be used before helper', async function (assert) {
```
MichalBryxi, line 23: Until some code fixing the addon is pushed, this test **will** fail. And that is ok as this PR is **only** intended to show the bug. I don't think I will have time anytime soon to fix the code itself, so very happy if someone can take-over and finish this.

gilest:
Thanks for providing this scenario. I've used it to stage a fix in #1094

PR #1090, opened 2025-01-24
Allow `@ember/test-helpers` v5.x, `tracked-built-ins` v4.x & `@glimmer/component` v2.x

mkszepp:
In out app we have some peer dependencies issues.

After this PR we whitelist the latest version of
- @ember/test-helpers (v5)
- tracked-built-ins (v4)
- @glimmer/component (v2)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,9 @@ jobs:
           - typescript@5.1
           - typescript@5.2
           - typescript@5.3
+          - typescript@5.4
+          - typescript@5.5
+          - typescript@5.6
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,11 +51,11 @@ jobs:
       fail-fast: true
       matrix:
         typescript-scenario:
-          - typescript@5.0
-          - typescript@5.1
-          - typescript@5.2
-          - typescript@5.3
-          - typescript@5.4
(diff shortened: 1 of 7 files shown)


SergeAstapov (approved):

SergeAstapov (commented):
looks great, unless TS <= 5.4 failures

mkszepp:
Yea... i have seen... i need only to understand which package makes this issue... i will look to find time... if you have any idea already let me know

mkszepp:
Affter testing a while... i found out whats the issue... `@tsconfig/ember` `3.0.8` works fine (which was used right now in addon), but `3.0.9` is not anymore compabile with old TS versions and so this tests were failed

mkszepp:
widen the support for test-waiters is maybe a future step because its not supported by test-helpers right now... https://github.com/emberjs/ember-test-helpers/pull/1524

but it isn't blocking this PR right now, because the issue of test-waiter isn't in app right now

SergeAstapov:
@gilest mind to take a look so we could merge and release?

gilest (approved):
Looks good to me. Thanks for updating the test matrix to cover recent TS versions.

I'd consider this an enchacment rather than a bugfix though 🤷🏻 

Want me to release after merge?

SergeAstapov:
> I'd consider this an enchacment rather than a bugfix though 🤷🏻

@gilest I'm always confused by such things, seen ppl do either bugfix and enhancement, so either way is good.

> Want me to release after merge?

@gilest please, if you don't mind to run the churn.

PR #1094, opened 2025-03-04
Create DEFAULT_QUEUE when FileQueueService is created

### LINKED ISSUE #1085: Service cannot be used before helper

gilest:
## Context

This issue occurs due to a _computation_ being rendered before a _dependency_ of that computation is updated – during the same render pass.

In this case:
```gjs
    const fileQueueService = this.owner.lookup(
      'service:file-queue',
    ) as FileQueueService;

    await render(
      <template>
        <ul>
          {{! `fileQueueService.files` is a computation of all files in all queues }}
          {{#each fileQueueService.files as |file|}}
            <li>
              {{file.name}}
            </li>
          {{/each}}
        </ul>

        {{! The `fileQueue` helper calls `fileQueueService.findOrCreate` which may create a new queue }}
        {{! Thus invalidating `fileQueueService.files` }}
        {{#let (fileQueue) as |queue|}}
          <label>
            <input type='file' {{queue.selectFile}} hidden />
            Select File
          </label>
        {{/let}}
      </template>,
    );
```

This is sometimes called a "backtracking re-render". Further reading:
- https://github.com/emberjs/ember.js/issues/13948

## Partial fix

A workaround to prevent this issue for the DEFAULT_QUEUE is to create it early, before the render pass.

However, with a named queue this would need to be done by consumers in their own app code. A demonstration of this has been added as a test.

Probably, some documentation should be updated to warn about this.

## Considerations

This is only a partial fix. As a breaking change we might want to restrict the `fileQueue` helper to only being able _find_ queues, and not create them.

That would mean users should create any named queues explicity in their own app code outside of rendering. This is a minor limitation, but I think it might be a better design since it suggests using autotracking in a more conventional way. Users would still be able to create dynamically named queues if they wanted, thought that's probably rare.

Partially fixes #1085

### DIFF
--- a/test-app/tests/integration/helpers/file-queue-helper-test.gts
+++ b/test-app/tests/integration/helpers/file-queue-helper-test.gts
@@ -5,7 +5,7 @@ import { render, click, settled, waitFor } from '@ember/test-helpers';
 import { DEFAULT_QUEUE, FileState } from 'ember-file-upload';
 import type { UploadFile } from 'ember-file-upload';
 import { selectFiles } from 'ember-file-upload/test-support';
-import { uploadHandler } from 'ember-file-upload';
+import { type FileQueueService, uploadHandler } from 'ember-file-upload';
 import { later } from '@ember/runloop';
 import fileQueue from 'ember-file-upload/helpers/file-queue';
 import { on } from '@ember/modifier';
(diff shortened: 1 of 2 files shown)


test-app/tests/integration/helpers/file-queue-helper-test.gts
```diff
@@ -274,4 +274,61 @@ module('Integration | Helper | file-queue', function (hooks) {
...
+
+    assert.dom('label').exists('regression would throw during rendering');
+  });
+
+  test('service can be used before helper with named queue, provided FileQueueService.create is called first', async function (assert) {
+    const fileQueueService = this.owner.lookup(
+      'service:file-queue',
+    ) as FileQueueService;
+    const customName = 'uploads';
+    fileQueueService.create(customName);
```
gilest, line 311: In the case of a named queue. A user would need to manually create it before a rendering pass to avoid this issue.

PR #1098, opened 2025-05-10
Remove unneeded peer

mkszepp:
@glimmer/tracking removed because it's a real package, but one we don't want to use. This comes up in embroider/vite where the presence of real packages always takes precedence over virtual packages. This is actually problematic because it can break reactivity in subtle ways, even if a dep graph is correct - allowing duplicates of dependencies, which for the glimmer internals, we don't want.

ref https://github.com/ember-cli/ember-addon-blueprint/pull/35

In additional we can remove safely also dependency ember-auto-import... its not necessary for v2 addon

### DIFF
--- a/ember-file-upload/package.json
+++ b/ember-file-upload/package.json
@@ -51,8 +51,7 @@
   "dependencies": {
     "@ember/test-waiters": "^3.0.0 || ^4.0.0",
     "@embroider/addon-shim": "^1.5.0",
-    "@embroider/macros": "^1.0.0",
-    "ember-auto-import": "^2.0.0"
+    "@embroider/macros": "^1.0.0"
   },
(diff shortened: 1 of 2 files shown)


SergeAstapov (approved):

SergeAstapov:
@mkszepp could you please update `pnpm-lock.yaml` file as CI fails with error

```
ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with <ROOT>/ember-file-upload/package.json
```

mkszepp:
@SergeAstapov now ci is green

SergeAstapov:
Thank you @mkszepp !

PR #1102, opened 2025-06-06
Remove reliance on ember-cli-mirage

beerinho:
`ember-cli-mirage` is an optional dependency and not actually required for the `uploadHandler` to function. As shown in the docs, the user will need to set up the Mirage server themselves for this to work at all but that server can be achieved through vanilla `miragejs`, as shown in this issue:  https://github.com/bgantzler/ember-mirage/issues/60

This PR removes the requirement on `ember-cli-mirage`.

### DIFF
--- a/ember-file-upload/src/mirage/upload-handler.ts
+++ b/ember-file-upload/src/mirage/upload-handler.ts
@@ -30,12 +30,7 @@ export function uploadHandler(
   fn: (this: void, db: unknown, request: FakeRequest) => void,
   options = { network: null, timeout: null },
 ) {
-  if (
-    macroCondition(
-      dependencySatisfies('miragejs', '*') &&
-        dependencySatisfies('ember-cli-mirage', '*'),
-    )
-  ) {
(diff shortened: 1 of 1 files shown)


jelhan:
The optional peer dependency on ember-cli-mirage should be removed as well, shouldn't it?
https://github.com/adopted-ember-addons/ember-file-upload/blob/2d7235b7c766221f6600b60d43269637f70491b6/ember-file-upload/package.json#L95

jelhan (approved):
Looks good to me.

@SergeAstapov was way more active than me in the last months and years in this repository. I like to give him the chance to review as well before merging.

gilest:
Hey @beerinho, thanks for your contribution!

I think you're right. We don't need to (and shouldn't) check for `ember-cli-mirage` since it's not the only or the blessed way to use Mirage with Ember.

I have checked the docs to see if they need updating and also searched for all mention of "mirage" in this repo, I think we're good to go.

Will release as an enhancement, since this is the removal of a check I don't think it's breaking.

PR #1135, opened 2026-01-06
Drop usage of @ember/runloop

gilest:
Marked as enhancement for a minor bump, but I could be convinced it should be breaking since there may be subtle timing changes due to the removal of runloop scheduling in DragListener.

Not sure how pertinent this work actually is anyway. The `eslint-plugin-ember` rules don't recommend using the runloop, but it's still public API. There's an exploring RFC about a new scheduling system for ember.

Generally speaking though, I don't think this addon needs to use any runloop functions, and removing them reduces the ember-specific API surface - a good thing for long-term maintenance.

### DIFF
--- a/ember-file-upload/src/system/http-request.ts
+++ b/ember-file-upload/src/system/http-request.ts
@@ -1,4 +1,3 @@
-import { bind } from '@ember/runloop';
 import RSVP from 'rsvp';
 import type { HTTPRequestOptions } from '../interfaces.ts';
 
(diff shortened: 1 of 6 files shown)


ember-file-upload/src/system/http-request.ts
```diff
@@ -83,39 +82,39 @@ export default class HTTPRequest {
       newPromise.then = promise.then;
       return newPromise;
     };
-    this.request.onabort = bind(this, function () {
```
gilest: It seems like `bind` in this file was being used for `this` binding. Replaced with arrow functions

NullVoxPopuli (approved):

PR #1177, opened 2026-07-08
Add recursive folder uploads via drag and drop

whatthewhat:
Adds an opt-in `@allowFolderDrop` argument to `FileDropzone`. When enabled, dropped directories are traversed recursively using the File and Directory Entries API (`webkitGetAsEntry`) — still the only cross-browser way to read dropped folders — and every contained file is added to the queue.

Each `UploadFile` now exposes `relativePath`, the file's location within the dropped directory (e.g. `reports/q3/deck.pdf`). It is also populated for files selected via `<input webkitdirectory>` (including through `queue.selectFile`) using the native `File.webkitRelativePath`.

Implementation notes:

- Entries are captured synchronously in the drop handler since browsers neuter `DataTransferItem`s once the handler yields.
- `readEntries` is called in a loop per directory — Chromium returns at most 100 entries per call.
- Unreadable files or subdirectories are skipped with a warning instead of failing the whole drop, and the dropzone's `active` state always resets via try/finally.
- Traversal is wrapped in `waitForPromise` so `settled()` waits for it in tests.
- `@filter` (and the `filter=` option of `queue.selectFile`) now receives the file's `relativePath` as a fourth argument, so whole subdirectories such as `__MACOSX/` can be excluded. Existing three-argument filters are unaffected.
- New `dragAndDropDirectory` test helper simulates nested folder drops.

Hidden files (e.g. `.DS_Store`) are intentionally not filtered; apps can exclude them by name or by `relativePath` with `@filter`.

Docs are updated (FileDropzone API, `UploadFile.relativePath`, file validation, testing helper), and the docsite demo has an "Allow folder drops" toggle for trying it in a real browser.

Based on the discussion in
https://github.com/adopted-ember-addons/ember-file-upload/pull/1066, extended with recursion and relative paths.

### DIFF
--- a/docs/api/class/upload-file.md
+++ b/docs/api/class/upload-file.md
@@ -84,6 +84,7 @@ An object which may be passed as an optional second argument to `upload()` and `
 | `rate`      | Current time in ms it is taking to upload 1 byte.                                                                           | `number` |
 | `state`     | The current state that the file is in.                                                                                      | `string` |
 | `source`    | The source of the file. This is useful for applications that want to gather analytics about how users upload their content. | `string` |
+| `relativePath` | Path of the file within the directory it was added from, including the directory name itself, e.g. `reports/q3/deck.pdf`. Populated for folder drops on a `<FileDropzone @allowFolderDrop={{true}}>` and for files chosen via an input with the `webkitdirectory` attribute. Empty string otherwise. | `string` |
 
 ## Methods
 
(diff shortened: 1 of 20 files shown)


whatthewhat:
Oops, sorry, I actually meant to open this against my fork (not 100% ready yet) 😬
That said, if there is still interest in adding this to ember-file-upload I'm happy to push it forward as this is something we'll need in our product.

